PR #14, opened 2017-02-07
Iss78

hszumila:
I added more variables to the tuple maker and TrackUtils in order to study hits shared between tracks and the effects.


omar-moreno (commented):
Given that there are several shared hit methods, they should be moved out to their own class at some point.  For now, this is fine since this class is a mess anyway.

omar-moreno (approved):

PR #20, opened 2017-02-08
loosened up v0 skim cuts

hszumila:
Loosened v0 skim cuts


baltzell (approved):
Looks fine.  One day *PMax should depend on beam energy.

PR #25, opened 2017-02-16
changes to tracking to resolve L0 not using L6

mrsolt:
Tracking is probably fixed.


JeremyMcCormick:
I'm not that familiar with the details of this but this XML seems very "hacky" ...

```
<driver name="TrackDataDriver" type="org.hps.recon.tracking.TrackDataDriver">
    <layerNum>7</layerNum>
</driver>
```

Why do you need to pass in a layer number to this Driver?

omar-moreno:
Yes.

omar-moreno (approved):
At some point we will change this to make it more generic, for now this is fine.

PR #32, opened 2017-02-21
Iss31

mrsolt:
adding track extrapolation variables to the tuple maker for layer 0-2.

### DIFF
--- a/analysis/src/main/java/org/hps/analysis/tuple/TupleDriver.java
+++ b/analysis/src/main/java/org/hps/analysis/tuple/TupleDriver.java
@@ -684,10 +684,10 @@ public abstract class TupleDriver extends Driver {
             }
 
             //////////////////////////////////////////////////////////////////////////
-            double rawHitTime[] = new double[7];
-            double rawHitTDiff[] = new double[7];
-            double rawHitMaxAmpl[] = new double[7];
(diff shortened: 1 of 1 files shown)


analysis/src/main/java/org/hps/analysis/tuple/TupleDriver.java
```diff
@@ -604,6 +609,19 @@ protected TrackState fillParticleVariables(EventHeader event, ReconstructedParti
             
             Track track = particle.getTracks().get(0);
             TrackState trackState = track.getTrackStates().get(0);
+            Hep3Vector extrapTrackPosL0;
+            Hep3Vector extrapTrackPosL1;
+            Hep3Vector extrapTrackPosL2;
+            if(trackState.getTanLambda() > 0){
+                extrapTrackPosL0 = TrackUtils.extrapolateTrack(track,topPos[0]);
```
omar-moreno, line 616: In the future, you should switch to using the extrapolator which makes use of the field map.  The analytic extrapolator should be fine for L1-L2 but I'm not certain it will give the most accurate results for L0.

omar-moreno (approved):
At some point, the field map should be taken into account in the extrapolation.  For now, using the analytic extrapolator is fine.

PR #46, opened 2017-04-14
removing tracking strategy requirement

mrsolt:
Removing the requirement of ele/pos having at least one of the same tracking strategy in order to make a V0 particle. It was a dumb requirement.


omar-moreno (approved):
Was it ...

PR #58, opened 2017-05-05
Remove users module

JeremyMcCormick:
Moved users dir to this github project:

https://github.com/JeffersonLab/hps-users


omar-moreno (approved):
It's about time.

omar-moreno:
@JeremyMcCormick when you inform users of the change, let them know that they need to do a 
`
mvn -DskipTests=true clean
`
before updating.  Otherwise, the users folder is not removed.

PR #77, opened 2017-05-17
Change charge transfer efficiency of layer 0 sensors.

omar-moreno:


omar-moreno:
@mrsolt Verified that this fixes the Layer 0 cluster multiplicity problem.  Below is the Layer 0 cluster multiplicity of tritrig events generated with the fix.
image

The change will be merged in.

PR #86, opened 2017-05-17
modified org.hps.evio.SvtEventFlagger

sebouh137:
Changed criteria for latency good flag:  was offset_time <= 27. changed
to offset_time <= 35.  This allows all 2016 events reconstructed
hereafter to have the good latency flag set.   (before, the good latency flag was only set for 2/3 of the events)

### DIFF
--- a/evio/src/main/java/org/hps/evio/SvtEventFlagger.java
+++ b/evio/src/main/java/org/hps/evio/SvtEventFlagger.java
@@ -44,6 +44,17 @@ public class SvtEventFlagger {
     private double nominalAngleTop = 0;
     private double nominalAngleBottom = 0;
 
+    private double maxGoodOffsetTime = 35;
+    private double minGoodOffsetTime = 0;
+    
+    public void setMaxGoodOffsetTime(double val){
+        this.maxGoodOffsetTime = val;
+    }
+    
+    public void setMinGoodOffsetTime(double val){
+        this.minGoodOffsetTime = val;
+    }
+    
     public void writeFlags(EventHeader event) {
         Date eventDate = getEventTimeStamp(event);
         if (eventDate != null) {
@@ -69,7 +80,7 @@ public class SvtEventFlagger {
 
         latencyGood = false;
         if (svtTimingConstants != null) {
-            if (svtTimingConstants.getOffsetTime() <= 35) {
+            if (svtTimingConstants.getOffsetTime() <= maxGoodOffsetTime && svtTimingConstants.getOffsetTime() >= minGoodOffsetTime) {
                 latencyGood = true;
             } else {
                 if (((event.getTimeStamp() - 4 * svtTimingConstants.getOffsetPhase()) % 24) < 16) {
--- a/evio/src/main/java/org/hps/evio/SvtEventFlagger.java
+++ b/evio/src/main/java/org/hps/evio/SvtEventFlagger.java
@@ -69,7 +69,7 @@ public class SvtEventFlagger {
 
         latencyGood = false;
         if (svtTimingConstants != null) {
-            if (svtTimingConstants.getOffsetTime() <= 27) {
+            if (svtTimingConstants.getOffsetTime() <= 35) {
                 latencyGood = true;
             } else {
                 if (((event.getTimeStamp() - 4 * svtTimingConstants.getOffsetPhase()) % 24) < 16) {
--- a/evio/src/main/java/org/hps/evio/SvtEventFlagger.java
+++ b/evio/src/main/java/org/hps/evio/SvtEventFlagger.java
@@ -47,6 +47,14 @@ public class SvtEventFlagger {
     private double maxGoodOffsetTime = 35;
     private double minGoodOffsetTime = 0;
     
+    public void setNominalAngleTop(double val){
+        nominalAngleTop = val;
+    }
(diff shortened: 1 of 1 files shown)


omar-moreno (changes requested):
1) This value should not be hard-coded.  It should be settable.  
2) Changing the hard-coded value is not the correct approach here.  This is not backwards compatible with the 2015 data and will change the events that have the latency flag set true.   Instead, as I mentioned, this should be settable.
3) How did you arrive at 35 ns?  It would be nice to link to some plots that shows this is a reasonable choice.

sebouh137:
1)  yes, I agree, this shouldn't be hardcoded.  What I think we should do
(eventually) is make SvtEventFlagger be a Driver in the recon, instead of
being buried deep inside of the EvioToLcio, so that the ranges of
acceptable offsets can be specified in the recon steering file.  But that
can be another issue that I or someone else can work on.

Just now I added setters for the range of acceptable values of the
offset_time in SvtEventFlagger (keeping my choice of 35 as the default for
the maximum acceptable offset_time if the setters aren't called), and am
going to make another pull request.

2)  The value that I have chosen and hardcoded is backwards compatible with
2015 data.  See the table on #3.  All the runs with the latency fixed in
the 2015 run have offset = 27.  Those that with the wrong latency value
have offset = 51 (with the exception of one run which has 75, because we
were testing whether increasing or decreasing the latency would fix the
problem).   The 2016 data has offset = 35.   So, requiring offset=<35 will
be backwards compatible with the 2015 and 2016 datasets.

3)  This is where I got the values from (Jeremy sent me these values which
are from the database):

+-----------+---------+---------------+--------------+-------------+


| run_start | run_end | collection_id | offset_phase | offset_time |


+-----------+---------+---------------+--------------+-------------+
|      4800 |    5037 |          1023 |            3 |          51 |
|      5076 |    5079 |          1023 |            3 |          51 |
|      5174 |    5216 |          1023 |            3 |          51 |
|      5236 |    5249 |          1023 |            3 |          51 |
|      5263 |    5298 |          1023 |            3 |          51 |
|      5310 |    5352 |          1023 |            3 |          51 |
|      5400 |    5528 |          1023 |            3 |          51 |
|      5558 |    5569 |          1023 |            3 |          51 |
|      5596 |    5598 |          1023 |            3 |          51 |
|      5686 |    5715 |          1023 |            3 |          51 |
|      5038 |    5074 |          1024 |            2 |          51 |
|      5139 |    5157 |          1024 |            2 |          51 |
|      5218 |    5233 |          1024 |            2 |          51 |
|      5251 |    5259 |          1024 |            2 |          51 |
|      5299 |    5309 |          1024 |            2 |          51 |
|      5375 |    5388 |          1024 |            2 |          51 |
|      5533 |    5555 |          1024 |            2 |          51 |
|      5575 |    5580 |          1024 |            2 |          51 |
|      5610 |    5639 |          1024 |            2 |          51 |
|      5642 |    5658 |          1024 |            2 |          51 |
|      5601 |    5602 |          1025 |            1 |          51 |
|      5603 |    5605 |          1026 |            5 |          51 |
|      5640 |    5640 |          1027 |            2 |          27 |
|      5641 |    5641 |          1028 |            2 |          75 |
|      5722 |    5777 |          1029 |            3 |          27 |
|      5779 |    5797 |          1029 |            3 |          27 |
|         0 |       0 |          1030 |            0 |           0 |
|   1000000 | 1000000 |          1030 |            0 |           0 |
|      7000 |  999999 |          1941 |            4 |          35 |
|      7455 |    7457 |          1974 |            4 |          35 |
|      7479 |    7496 |          1974 |            4 |          35 |
|      7403 |    7411 |          1974 |            4 |          35 |
|      7628 |    7644 |          1974 |            4 |          35 |
|      7795 |    7809 |          1974 |            4 |          35 |
|      7946 |    7962 |          1974 |            4 |          35 |
|      7966 |    7982 |          1974 |            4 |          35 |
|      8072 |    8100 |          1974 |            4 |          35 |
|      7373 |    7374 |          1979 |            3 |          35 |
|      7415 |    7415 |          1979 |            3 |          35 |
|      7447 |    7453 |          1979 |            3 |          35 |
|      7611 |    7614 |          1979 |            3 |          35 |
|      7653 |    7653 |          1979 |            3 |          35 |
|      7779 |    7786 |          1979 |            3 |          35 |
|      7794 |    7794 |          1979 |            3 |          35 |
|      7963 |    7965 |          1979 |            3 |          35 |
|      7983 |    7989 |          1979 |            3 |          35 |
|      8024 |    8059 |          1979 |            3 |          35 |
|      7460 |    7461 |          1980 |            2 |          35 |
+-----------+---------+---------------+--------------+-------------+

JeremyMcCormick:
I think maybe this PR should be taken down until we decide what we want to do about this.  It seems that maybe we will need a condition here so that the SVT event flagger can be configured properly for 2015 and 2016 runs.  I don't like the idea of putting those numbers in a steering file, as this value is dependent on run numbers.

JeremyMcCormick:
I talked with Omar and we will close this for now...

We are moving away from even using the latency flag for the 2016 data, so this modification is not necessary.

Also, you can't just hard-code a latency value here and assume it is correct.  This requires a lot more analysis to justify.

PR #91, opened 2017-05-24
Iss72: Conditions system cleanup

JeremyMcCormick:
Implements the following changes to the conditions system:

- Removes the unneeded XML configuration file from the database conditions manager

- Changes test logging level to INFO so conditions info is printed

- Simplifies the connection management.  It is assumed now that usage of the conditions manager outside of the `Driver.detectorChanged()` method requires the user to explicitly reopen and then close the database connection.  The connection is no longer closed and reopened when converting each collection.

- Adds the `ConditionsIT` integration tests to the integration-tests module which runs a number of different job and data configurations to test that the correct conditions are loaded.

- Simplifies conditions system initialization.  Now creating a new `DatabaseConditionsManager` will automatically install it as the global/default conditions manager.  Calling the `getInstance()` method will only return this instance now, not perform any initialization.

- Minor comment and log message changes in the db manager

- Fixed conversion of test run data to LCIO because of error in SVT data converter (can be changed or reverted later but it was needed so that Test Run 2014 data could be loaded for conditions testing)


omar-moreno:
Does this need to be merged in before I start working on the conditions system fixes?

JeremyMcCormick:
> Does this need to be merged in before I start working on the conditions system fixes?

No, you should be fine.

JeremyMcCormick:
Going to take this down until I test further.

mholtrop:
Tested with:
`java  -cp /data/HPSlcsim/merge_conditions/distribution/target/hps-distribution-4.0-SNAPSHOT-bin.jar org.hps.evio.EvioToLcio  -x /org/hps/steering/recon/EngineeringRun2015FullRecon.lcsim  -r -d HPS-EngRun2015-Nominal-v4-4-fieldmap -R 5772 /data/HPS/engrun2015/evio/hps_005772.evio.0 -DoutputFile=/data/HPS/data/recon2_005772_0.slcio -n 10000
`
And, with and without internet connection with:
`java -Dorg.hps.conditions.connection.file=local.prop  -Djava.util.logging.config.file=logging_local.properties  -cp /data/HPSlcsim/merge_conditions/distribution/target/hps-distribution-4.0-SNAPSHOT-bin.jar org.hps.evio.EvioToLcio  -x /org/hps/steering/recon/EngineeringRun2015FullRecon.lcsim  -r -d HPS-EngRun2015-Nominal-v4-4-fieldmap -R 5772 /data/HPS/engrun2015/evio/hps_005772.evio.0 -DoutputFile=/data/HPS/data/recon2_005772_0.slcio -n 10000`

No difference seen from "master".

Issue #92, opened 2017-05-24
set vertex position to -5.0 in MakeTuples2016.lcsim


sebouh137:

normangraf:
I'd like to see an analysis note documenting how this number was derived. @bloodyyugo @baltzell

JeremyMcCormick:
We don't know that the target is at -5 mm for 2016 data.  

@normangraf is working on this and changes will be submitted under a different issue branch.

Closing for now.

PR #93, opened 2017-05-24
added <beamPosZ>-5.0</beamPosZ> into MakeTuples2016.lcsim

sebouh137:


omar-moreno:
Before this is approved, I want to understand how you concluded that it needs to be at -5 mm.

normangraf:
I agree. See my [comment ](https://github.com/JeffersonLab/hps-java/issues/92#issuecomment-304098111) on the issue.

sebouh137:
I was attempting to make it consistent with what was being used for 2015.  That's where the -5 mm comes from.

omar-moreno:
Norman has already shown that this isn't the case for 2016.

normangraf:
Plotting the unconstrained vertex z position for Møller candidates in the 2016 data does not result in a distribution peaked at -5. How was this number derived for the 2015 data and where is it documented? Before we proceed to reprocess the 2015 data and process the 2016 data I'd like to have a clear understanding of how this number was derived and see that process applied to both data sets.

PR #109, opened 2017-06-06
added AmbiguityResolver in hps.recon.tracking

mdiamon:
also added a HoleCreationDriver and Tests under my hps-users

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/AcceptanceHelper.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/AcceptanceHelper.java
@@ -1,0 +1,212 @@
+package org.hps.recon.tracking;
+
+import hep.physics.vec.Hep3Vector;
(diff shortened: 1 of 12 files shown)


JeremyMcCormick (changes requested):
* Move tests to `hps-java/tracking/src/test/java`.  The users module no longer exists (it was moved to a separate git repo).

* Probably we should upload your test files into the http test directory if you want them to actually run for other people.  If you put these files someplace accessible to me on SLAC AFS or NFS I can do this for you.  Then you can use the file cache to download them (see other existing test cases for example of doing this).

* Minor style point: You don't really need underscores in front of class variables in Java and we don't tend to do this.  Instead you can use the `this` keyword to disambiguate (like the `this` or `self` keywords in C++ and Python).

JeremyMcCormick (approved):

omar-moreno (approved):

PR #110, opened 2017-06-07
Fix convention of module number to equal the java geometry.

perhansson:


omar-moreno:
So are we merging this in?

JeremyMcCormick:
Going to close and submit using another branch.

PR #112, opened 2017-06-07
Iss87: Updates for running recon on LCIO files

normangraf:
Created several new classes needed for reconstructing events from lcio files and tests for them.
Added toString methods to a few existing classes.
Main change is to TrackDataDriver, which merits a closer look from the reviewer(s).

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/TrackDataDriver.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/TrackDataDriver.java
@@ -342,14 +342,14 @@ public final class TrackDataDriver extends Driver {
                 
                 // Create a new TrackData object and add it to the event
                 TrackData trackData = new TrackData(trackerVolume, trackTime, qualityArray);
-                System.out.println(trackData);
+//                System.out.println(trackData);
                 trackDataCollection.add(trackData);
                 trackDataRelations.add(new BaseLCRelation(trackData, track));
 
                 // Create a new TrackResidualsData object and add it to the event
                 TrackResidualsData trackResiduals = new TrackResidualsData((int) trackerVolume, stereoLayers,
                                 trackResidualsX, trackResidualsY);
-                System.out.println(trackResiduals);
+//                System.out.println(trackResiduals);
                 trackResidualsCollection.add(trackResiduals);
                 trackToTrackResidualsRelations.add(new BaseLCRelation(trackResiduals, track));
             }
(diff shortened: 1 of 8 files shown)


tracking/src/main/java/org/hps/recon/tracking/TrackDataDriver.java
```diff
@@ -239,27 +247,58 @@ protected void process(EventHeader event) {
...
+                    // totalT0
+                    // totalHits
+                    // trackerVolume (0 == top, 1 == bottom)
+                    //
+                    // Can we do this without resort to HelicalTrackCross?
+                    for (Object o : rotatedStereoHit.getRawHits())
+                    {
+                        RawTrackerHit rth = (RawTrackerHit) o;
+//                        System.out.println(rth);
+                        totalT0 += rth.getTime();
```
omar-moreno, line 275: Before, the totalT0 was using the cluster time which is defined as the amplitude weighted time of the raw hits that compose it.  Now it's using the raw hit times so the definition has changed.  I don't think this is being used in the recon, but this might impact some analyses.

JeremyMcCormick (commented):
Minor issue: Format this code so that opening brackets are on the same line.  This is not being done consistently in this code.

JeremyMcCormick:
> Main change is to TrackDataDriver, which merits a closer look from the reviewer(s).

I'm assuming you also checked the evio2lcio recon procedure to double check that your changes didn't break anything for 2015/2016 data?  

If that works probably your changes to this Driver are okay.

normangraf:
I've run events from scratch (evio2lcio) and see no issues with proceeding with this pull request.

omar-moreno:
Did you see my comment on the t0 sum definition?

normangraf:
I missed that comment. What is the issue?

omar-moreno:
I left some details on the pull request page.

tracking/src/main/java/org/hps/recon/tracking/TrackDataDriver.java
normangraf: Thanks for noticing this. I'll double-check on any differences in the totalT0 reported by both methods and modify if necessary.

normangraf:
I've confirmed that using the raw times for the track time is incorrect. I've also found an offset (~2ns) when plotting the track time as determined from the cluster times.

JeremyMcCormick:
This is old and now has conflicts with master.  I'm going to close but please reopen if you want to merge some future version of this branch.

PR #120, opened 2017-06-13
fix MergeTrackCollections

mdiamon:
even though no one actually uses the PartialTracks collection for anything, MergeTrackCollections has to create this collection or else the dst-maker complains.

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/MergeTrackCollections.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/MergeTrackCollections.java
@@ -30,26 +30,34 @@ import org.lcsim.util.test.TestUtil.TestOutputFile;
 public class MergeTrackCollections extends Driver {
 
     private String outputCollectionName = "MatchedTracks";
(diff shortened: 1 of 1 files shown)


JeremyMcCormick (approved):

omar-moreno:
We should have just removed checking for partial tracks in the DST maker instead of adding them back into the recon ...

Actually, I would have preferred to leave them out because at the moment, the DST maker checks for the collection in order to make sure it's not written to the DST

```
        // Don't write partial tracks to the DST.  Partial tracks are tracks
        // whose hits are a subset of another track in the event.
        if (tracks == (EVENT::LCCollection*) event->getCollection(PARTIAL_TRACKS_COL_NAME)) continue;
```

Anyway, I'll probably just remove this check from the DST maker.

omar-moreno:
@JeremyMcCormick Can you revert this merge.

mdiamon:
I pushed the branch issue119 again, with the partialTracks taken back out

PR #121, opened 2017-06-14
take PartialTracks back out

JeremyMcCormick:


omar-moreno (approved):
The DST maker will no longer check for the partial tracks collection.

PR #132, opened 2017-06-22
Iss130

normangraf:
Fixed some major errors in logic.
This is a patch which is only being applied to the Unconstrained vertex collections.. The underlying problem still needs to be fixed.


omar-moreno (approved):
I verified that these changes fixes the null pointer exception being reported on the master.

PR #137, opened 2017-06-28
Copy FastCheck and other classes from lcsim

mdiamon:
The three-point helix check was found in FastCheck.class, in the lcsim library.  I ported it into hps as FastCheck.java.  The FastCheck was included by another lcsim class as a private member (with no public set methods), which was in turn privately included by another class, and so on... meaning I had to port several other lcsim classes over as well, in order to propagate the use of FastCheck.java all the way up the chain.


omar-moreno:
Let's wait to merge this until after we make the release for pass1.

PR #144, opened 2017-07-06
Issue134: Update tracking unit tests

mdiamon:
Unit tests in recon/tracking: deleted some outdated ones, updated some, added some


JeremyMcCormick:
Which of these tests runs locally only and which ones use remote test data?

Ideally, these should all run on remote test data.  Otherwise, the tests need to be excluded from the build as they will fail due to missing data that only you have locally.

mdiamon:
Ah, they can run on any data, but the versions I pushed have local data.  So you're saying I should change them to remote data.  OK, next week I should probably discuss with some other people (Norman? Omar?) which remote data should be in there by default.


-- Miriam


normangraf:
Tests should use remote files by default. (Note that one should write the test cases such that they check first to see whether the files have already been downloaded and cached). There should already be files available for most of our testing purposes, but we should review what those files are and identify any new ones that might be needed.?


JeremyMcCormick:
Yes, all tests should use remote test data so that others can run them.

There's no issue adding files to our web dir for your tests if appropriate ones do not exist there already.  You would just need to point me to the current file locations so I can copy them.

JeremyMcCormick (changes requested):
- Change `tracking/pom.xml` to exclude tests that should not run; also remove the general exclude which is filtering out all tracking tests

- Change all tests to use remote data and do not return if `inputFileName` is null (remove that code)

- Fix all broken tests

```
Results :

Failed tests:
  TruthResidualTest.testRecon:51 Mean of layer 1 x truth hit residual is not zero 0.00817456119323906
Tests in error:
  ComparisonTest.testRecon:47 ? IllegalArgument The path Tracks per Event does n...
  MergeTrackCollectionsTest.initializationError ? ExceptionInInitializer
  HpsGblRefitterTest.testHpsGblRefitter:42 ? Runtime java.lang.ClassCastExceptio...
```

- Delete `HpsGblRefitterTest` as it is broken and does not test anything

- Do not write output files to the `tracking` base dir.  Instead use `TestOutputFile` to automatically write them to `target/test-output` e.g.

```
File outputFile = new TestOutputFile("myTestOutputFile");
```

If you are using classes that need a file path string instead of a Java File then use `target/test-output/filename` instead.

JeremyMcCormick (approved):

PR #145, opened 2017-07-06
Iss118: Performance optimizations for raw tracker hit pulse fitting

sebouh137:
The "a" stands for a more "aggressive approach" to improving pulse fitting than the original iss118 branch has.  Basically, calculates the estimated amplitudes for all the samples in a single function, in such a way that it reduces the number of calls to Math.exp().  This is in addition to the improvements that are in branch 118.

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/PulseShape.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/PulseShape.java
@@ -71,6 +71,8 @@ public abstract class PulseShape {
             peak_t = 3.0 * Math.pow(tp * Math.pow(tp2, 3), 0.25); //approximate solution to exp(x)=1+x+x^2*tp/(2*tp2), where x=(1/tp2-1/tp)*t
             peak_amp = getAmplitudeIntegralNorm(peak_t);
             
+            //System.out.println(tp + "  " + tp2);
+            
             A = (Math.pow(tp, 2) / Math.pow(tp - tp2, 3));
             B = (tp - tp2) / (tp * tp2);
         }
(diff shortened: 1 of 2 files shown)


JeremyMcCormick:
@sebouh137 Please provide detailed evidence such as before/after plots that your changes give the same results without the optimizations.

omar-moreno:
HI @sebouh137, before merging these changes, some basic hit level studies need to be done.  I would start by checking that you are getting the same number of hits as before and then looking at the fit results in detail (e.g. chi2 distribution, t0 distribution).

sebouh137:
[issue118.pdf](https://github.com/JeffersonLab/hps-java/files/1145320/issue118.pdf)
The attachment is the talk i gave yesterday (Wednesday, July 12) at the Software meeting.  I have also included a plot of hit time versus difference in hit time.  Almost all of the hits for which the two methods disagree (which constitute <.1% of the total hits) are after 50 ns, well after our trigger window.  I believe what is happening is that for the later hits, there are fewer samples after the beginning of the pulse that remain to be fit.

mdiamon (commented):
[ap_prompt-iss118.pdf](https://github.com/JeffersonLab/hps-java/files/1206015/ap_prompt-iss118.pdf)
[hps_005772-iss118.pdf](https://github.com/JeffersonLab/hps-java/files/1206016/hps_005772-iss118.pdf)

Looks good to me.  Locally, I merged it into the latest master, and there were no conflicts.  I then compared results with it to reference master results, for 10K run 5772 data events and for 1700 prompt A' MC events.  Plots are essentially identical.

normangraf (approved):

PR #147, opened 2017-07-06
Iss146: Track-cluster matching improvements

sebouh137:
Copied everything from the dev-tc-match branch.


JeremyMcCormick:
There are a lot of changes here, @sebouh137.

Please describe exactly what you changed in the track-cluster matching.

Also provide plots or other data that shows old vs new results and verifies that you didn't break anything.

sebouh137:
I will present the results on Tuesday at the analysis meeting.  I prepared
a presentation thinking I would be able to present it at either the
software or tracking meeting (but both were cancelled).

sebouh137:
here is the presentation I gave on tuesday. July 11 (name of file is a misnomer):  
[070317.pdf](https://github.com/JeffersonLab/hps-java/files/1146120/070317.pdf)

baltzell:
If beam energy is not 1.05 or 2.306 GeV (+/- 0.1), then no matching will happen?  Why not just let it use 2.3 GeV parameters for 4 and 6 GeV for now (if E<2, use these, else use those), instead of turning off all matching for high energy simulations, especially since we know 1.2 parameters really weren't too bad for 2.3 for real data.

Also, if not moving those parameters to conditions db, they should be declared static final.

sebouh137:
I agree, Nathan.  I will change the code so that it uses 2.3 GeV parameters if the beam energy is anything greater than 2 GeV.  After the zeroth pass of our 4 GeV or 6 GeV runs (during which we'd use the 2 GeV parameters), we can change the code to add additional sets of parameters for the 4 or 6 GeV beam.

normangraf:
?Hi Sebouh,


We should at some point move these fit parameters to the conditions database. Then we we can more easily adjust run ranges, etc.

Norman


normangraf (approved):
This pull request requires the comparison histograms for the integration tests to be rebuilt.

PR #149, opened 2017-07-06
 RewriteFlagsDriver was removed from MakeTuples2016.lcsim

rafopar:
ntuplemaker was crushing with a new release, while with a tpass1c it didn't crush.
It was trying to call a driver from users directory
org.hps.users.holly.RewriteFlagsDriver
Sebouh mention that users were completely removed from the 3.11 release.

As Sebouh suggested, RewriteFlagsDriver is removed from MakeTuples2016.lcsim,
now it seems to work again


JeremyMcCormick (approved):

rafopar:
Jeremy, could you make a new release? and I will run the pass using the new release?

PR #164, opened 2017-07-19
Committing Code for Pulse Fitting Cosmic

LucaMarsicano:
Implemented pulse fitting for Ecal calibrations with cosmic rays.


hszumila:
I think the comments at the top of the code are the same as from the codes they were copied from. At a minimum, this should have the you as the author, the function of the code and its output. Everything else looks fine.

hszumila (approved):

PR #169, opened 2017-07-26
Iss168: Add fieldmap URLs to compact files and simple test case

JeremyMcCormick:
- Added url attribute to fieldmap detectors

- Added simple test that makes sure fieldmap data is accessible for relevant detectors

- Updated lcsim version to 4.0-SNAPSHOT to pull in changes to `FieldMap3D` class for file caching support

### DIFF
--- a/detector-data/detectors/HPS-EngRun2015-1_5mm-v3-4-fieldmap/compact.xml
+++ b/detector-data/detectors/HPS-EngRun2015-1_5mm-v3-4-fieldmap/compact.xml
@@ -857,7 +857,10 @@
     </readouts>
 
     <fields>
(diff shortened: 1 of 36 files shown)


JeremyMcCormick:
Don't merge this yet.  I'm still testing at JLAB.

JeremyMcCormick:
@mholtrop 

Regarding offline usage, as discussed in today's software meeting, I have checked this in the following way...

I login to jlabl5 and then run this command from my hps-java dir:

```
mvn test -Dtest=FieldMapTest -Dorg.lcsim.cacheDir=$PWD
```

This caches the field map data for all relevant detectors into the current directory.

Then I login to ifarm and I run the same test (the `-o` switch is making sure Maven doesn't try to download any deps from the internet):

```
mvn -o test -Dtest=FieldMapTest -Dorg.lcsim.cacheDir=.
```

I can see in this test that it is using and finding all the field map data files downloaded from previously running the test on jlabl5.

The mechanism used is the exact same one that would be activated in recon or other jobs as it is built into the FieldMap3D class within lcsim.

Is this sufficient for testing that offline mode works okay?

normangraf:
Looks good.

normangraf (approved):

PR #200, opened 2017-08-08
Iss82: SVT alignment constants must be explicitly enabled when building detector

JeremyMcCormick:
The system property `org.hps.conditions.enableSvtAlignmentConstants` must be set to enable reading of SVT alignment constants from the database.  Otherwise, the constants from the compact.xml file are used.

This replaces the `disableSvtAlignmentConstants` property which no longer has any effect.


mholtrop:
Seems to work fine in tests and running a job on run 5772.

PR #206, opened 2017-08-19
Iss167: Change target position to z=0 in MC steering file

brigidcrowe:
Set target position to z=0 in MakeTuplesMC steering files.


JeremyMcCormick:
@byale-jlab Is this just bringing this particular steering file up to date with changes that have already been made to other MC steering files?

PR #210, opened 2017-09-04
Improved track extrapolation

mrsolt:
Extrapolated track to each sensor (instead of just the layer). This is compatible with L0 detector.


JeremyMcCormick:
The TupleDriver doesn't seem like the appropriate place to extrapolate the track, especially if this has been done elsewhere already.  Wouldn't you want what's written out by this driver to match the previous extrapolation rather than redo it here?

omar-moreno (approved):
This looks fine.

PR #220, opened 2017-09-20
Iss83: Implement reconstruction ITs

normangraf:
Please leave this issue open, as more development work is expected.


JeremyMcCormick (approved):

JeremyMcCormick:
I was able to run these successfully from a clean build (I also removed the old AIDA ref file from my cache dir beforehand if that matters).

It would be nice to have a confirmation from one or 2 more people that these work for them before merging.

Issue #251, opened 2017-11-24
Fix Recon getting stuck on specific events.


mholtrop:
When running recon, the process gets stuck on specific events. The files for which this happens from run 5772 are documented here: [tpass7f]( https://confluence.slac.stanford.edu/display/hpsg/tpass7f)

Specific event:run 5772,  file 240, event number 61402770

Ouput:
```
2017-11-24 12:24:11 [INFO] org.lcsim.job.EventPrintLoopAdapter recordSupplied :: event: 61402766; time: 1431857544478647336; seq: 0
2017-11-24 12:24:11 [INFO] org.lcsim.job.EventPrintLoopAdapter recordSupplied :: event: 61402767; time: 1431857544478724340; seq: 1
2017-11-24 12:24:11 [INFO] org.lcsim.job.EventPrintLoopAdapter recordSupplied :: event: 61402768; time: 1431857544478732220; seq: 2
2017-11-24 12:24:11 [INFO] org.lcsim.job.EventPrintLoopAdapter recordSupplied :: event: 61402769; time: 1431857544478816276; seq: 3
```

At which point the process is stuck forever until the process is killed.

mholtrop:
Stack trace:
```
Evio2Lcio [Java Application]	
	org.hps.evio.EvioToLcio at localhost:62943	
		Thread [main] (Suspended)	
			TrackUtils.extrapolateTrackUsingFieldMap(TrackState, double, double, double, FieldMap) line: 1566	
			TrackDataDriver.process(EventHeader) line: 276	
			TrackDataDriver(Driver).doProcess(EventHeader) line: 261	
			Driver.processChildren(EventHeader) line: 271	
			Driver.process(EventHeader) line: 187	
			DriverAdapter.recordSupplied(RecordEvent) line: 74	
			JobManager(JobControlManager).processEvent(EventHeader) line: 819	
			EvioToLcio.run() line: 618	
			EvioToLcio.main(String[]) line: 92	
		Daemon Thread [Abandoned connection cleanup thread] (Running)	
	/Library/Java/JavaVirtualMachines/jdk1.8.0_92.jdk/Contents/Home/bin/java (Nov 24, 2017, 12:29:48 PM)	
```

Code:
```
   while (currentPosition.x() < endPositionX) {

            // The field map coordinates are in the detector frame so the
            // extrapolated track position needs to be transformed from the
            // track frame to detector.
[...]

PR #252, opened 2017-11-25
Iss251 - Fix Recon getting stuck on specific events. 

mholtrop:
See comments on issue #251 - the problem was extrapolation to the ECAL for tracks that would never make it that far.


normangraf:
Was this branched off of the master? It is creating an hps-distribution-4.0-bin.jar file instead of an hps-distribution-4.1-SNAPSHOT-bin.jar .

normangraf:
I've run the integration tests and they all passed successfully, so except for the jar file naming (see previous comment) I think we are good to go for the merge. @JeremyMcCormick @mholtrop

mholtrop:
Hi Norman,

I based this on the hps-java-4.0 release, since I saw this as a patch to the release and I did not want to mix in other changes to release in testing it. It should go into the 4.1 release that we run before the full pass7. Merging with master should not pose an issue.

Best,
	Maurik

normangraf (approved):

PR #254, opened 2017-11-27
added steering file V0CandidateFilter2015Pass7.lcsim

sebouh137:
Added the steering file.  I tested it by plotting the pair momenta of v0s, requiring only that there be good svt flags, and that the tracks be on opposite sides of the detector for both the original file and the skimmed file, and the two histograms are identical.


JeremyMcCormick:
Is this still needed/wanted?  

The single filter Driver in this config is now part of the 2015 recon chain I believe (via a different PR).

PR #268, opened 2017-12-06
Iss243: eliminating root cause of Matrix Is Singular errors

mdiamon:
Rather than just catching the error and discarding the track (as a patch), eliminates the source of the problem.  Slightly changes how high-d0/low-momentum tracks are handled.


normangraf (approved):
Changes causes a few tracks to be lost, but within the tolerance. Comparison histograms will have to be regenerated.

PR #272, opened 2017-12-08
Iss266: Fix TrackState at ECal

mdiamon:
Now uses TrackState at final sensor (if available), and starts the extrapolation to the ECal from the helix intercept at final sensor.


normangraf:
Has the effect of this change on the track-cluster matching been studied? In particular, are the residuals still centered on zero and will sigmaX and sigmaY have to be changed for the track-cluster nSigma calculation? @sebouh137 @baltzell

mdiamon:
As far as I know, effects on track-cluster matching have not been studied at all.

PR #292, opened 2018-02-07
Eliminate re-vertexing from tuple: Iss286

mdiamon:
Branch 286b
Also features some re-factoring of the tuple code, into abstract Maker classes (for data, MC, and full truth MC) and concrete Driver classes.


mrsolt (approved):
Looks good.

PR #300, opened 2018-02-10
Fix missing ID information in inactive Si readout

JeremyMcCormick:
Made small patch to the detector converter code to add an ID spec based on the readout name.

Includes regenerated LCDD file for v7 detector.

Verified that regenerating the data using the new detector writes valid ID information; I can see non-zero system, layer and module values in the LCSim Event browser in JAS3 now.


mrsolt (approved):
Looks good.

PR #303, opened 2018-03-13
Fix dumb conditions init bug.

JeremyMcCormick:
Fix conditions initialization bug when starting monitoring app.

Checked after this fix that the following configuration executes successfully on EVIO data from run 5772:

```
#Tue Mar 13 15:47:26 PDT 2018
SteeringType=RESOURCE
DisconnectOnEndRun=true
AIDAServerName=hps-monitoring-app
WaitTime=1000000000
StationName=MY_STATION
StationPosition=1
LogLevel=ALL
EventBuilderClassName=org.hps.evio.LCSimEngRunEventBuilder
MaxEvents=-1
WaitMode=TIMED
ChunkSize=1
ConditionsTag=engrun2015_pass7
DisconnectOnError=true
DetectorName=HPS-EngRun2015-Nominal-v5-fieldmap
SteeringResource=/org/hps/steering/monitoring/SVTMonitoring.lcsim
EtName=ETBuffer
Verbose=false
Port=11111
LogToFile=false
DataSourceType=ET_SERVER
Host=localhost
QueueSize=0
LogLevelFilter=ALL
Prescale=1
ProcessingStage=LCIO
Blocking=false
```

### DIFF
--- a/monitoring-app/src/main/java/org/hps/monitoring/application/util/ResourceUtil.java
+++ b/monitoring-app/src/main/java/org/hps/monitoring/application/util/ResourceUtil.java
@@ -36,7 +36,15 @@ public final class ResourceUtil {
      * @return the list of available conditions tags
      */
     public static String[] getConditionsTags() {
-        return DatabaseConditionsManager.getInstance().getAvailableTags().toArray(new String[] {});
+        // FIXME: New database manager should probably not be instantiated here.
+        DatabaseConditionsManager mgr = new DatabaseConditionsManager();
+        String[] tags = mgr.getAvailableTags().toArray(new String[]{});
+        try {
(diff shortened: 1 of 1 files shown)


mholtrop (commented):
Seems simple enough, however, why the comment: FIXME: New database manager should probably not be instantiated here.
If that is the case, why not fix this properly, since no one will likely read the FIXME?

JeremyMcCormick:
I would just merge it because it fixes a serious bug that prevents the
monitoring app from loading...

The bigger issue would require some rearchitecting of the monitoring app
which I did not want to implement.

mholtrop (approved):
Tested this same fix on #304 and it worked there.

PR #326, opened 2018-09-26
MOUSE

mdiamon:
Object Standardization


mdiamon:
Combined with #328

JeremyMcCormick:
You should setup your StandardCuts class in the detectorChanged() or startOfData() methods instead of in the setters.  Beam energy will have been set by then so you don't need to do this in multiple places.

mholtrop:
Passed all the tests. Yay!

mholtrop (approved):
All tests passed, seems good to go.

PR #344, opened 2018-11-09
Added target and beamspot positions to 2016 recon steering file.

normangraf:


mholtrop (approved):

mholtrop:
Just realized that the integration tests would not be using the new detector, so they may not be a good test of this merge. Still seems OK to go ahead and have the test pass2 be the full verification.

PR #346, opened 2018-11-15
Millepede-readable binary file from GBL

mdiamon:
Does not change any standard output. Just adds the ability to produce millepede binary file and ROOT plots file directly from GBL, for alignment.


mholtrop:
This seems OK.
Only comment: The HpsGblRefitter is no longer a drive, unlike the comment says.

mholtrop (approved):
Passes the integration tests. I am not in a position to test whether the GBL output is correct.

PR #362, opened 2018-11-29
Iss40

mholtrop:
This pull only affects the specific new geometry for the Hodoscope. It needs to be merged before we can merge iss166, which is a major update of the readout code in support of the Hodoscope.


omar-moreno (approved):
All of these files are new additions to hps-java and don't impact the master in any way.  It's fine to merge.

PR #372, opened 2018-12-12
Iss361

rafopar:
Some of MOUSE cuts were set in the steering files, which basically loosen MOUSE cuts.

In order to set chu2/NDF cut, the driver GBLRefitterDriver.java has modified, i.e.
two separate methods are added for 5 and 6 hit tracks.


mholtrop (approved):
OK.

PR #373, opened 2018-12-12
Issue 371 Resolution

cbravo135:
Changes SVT pulse time minimization to use migrad.


cbravo135:
simpvmig

cbravo135:
simpvmigerrors

normangraf:
@mholtrop , I'm OK with the results shown here in terms of t0 performance, however it would be good if you could run some timing test as well before approving this PR.

omar-moreno:
It would great the get the profiling done ASAP so this can be used in
pass2.

mholtrop:
Hi Norman,

Yes, and I am sorry for this opinion, but this it is horrible.

I really think it is not a good idea to spend 31% of the total CPU time for analyzing data on fitting the SVT hits. That is up from the already really horrible 14%. Same is true for the ECAL hits, which were at 21%. Yes, a total of 1/2 the CPU time is spend converting raw hits to pulses.

Old profile for master: http://nuclear.unh.edu/HPS/Profiles/Call_Tree_doProcess_2018_12_11.xml <http://nuclear.unh.edu/HPS/Profiles/Call_Tree_doProcess_2018_12_11.xml>
New profile for iss371: http://nuclear.unh.edu/HPS/Profiles/Call_Tree_doProcess_2018_12_13-iss371.xml <http://nuclear.unh.edu/HPS/Profiles/Call_Tree_doProcess_2018_12_13-iss371.xml> 

I have real doubt that this process was ever optimized for CPU consumption, but let’s not make it worse unless someone can convince me that this really makes a difference.

If it DOES make a difference, then OK, we’ll take the hit on extra CPU.

We will not be able to do this for the 2019 data set. If we cannot really significantly speed up this part of our code, I will rewrite the EVIO to LCIO conversion step, including the raw pulse fitting, as a separate step, probably in multi-threaded C++, and we run that step once over the data, and be done with it.

Best,
	Maurik

normangraf:
Hello Maurik,


No need to apologize. This has been a long-festering issue that we should now address in earnest. We clearly can not afford to do this with the 2019 data.


Norman


baltzell:
The last profiling I remember seeing had ecal pulse fitting at at most 10%, so didn't seem like a big target for optimization at the time.  But 20% is getting up there.  Hmm, looking through the archives, I guess this is the first profiling on 2016 data:

https://confluence.slac.stanford.edu/display/hpsg/Reconstruction+Code+Profiling

mholtrop:
Hi Nathan,

At 10% for ECAL and another 10% for SVT, I was only a little bit concerned, but at the 16% for ECAL and 31% for SVT (for issue371), we are spending too much time fitting pulses. If this is truly needed, then we should do this as a separate step, perhaps right when the data comes in, and then not repeat this fits with every pass.
I think there are two reasons for the increased percentage: 1) The MOUSE cuts make the tracking code take a lower fraction of the overall CPU, because tracks are cut earlier in the process. This is a really good thing.  2) Changing from simplex to migrad makes the SVT cuts run much slower. You can see the difference in the two profiles. 

I don’t think this is a trivial thing to fix though. We would like good timing on the pulses. I just think that we should consider being more clever about it. I am starting to work with a very clever undergraduate student who is good with code, and trying to see if this is something he would like to attack.

Best,
	Maurik

mdiamon:
Maurik is correct that the tracking code is now taking less CPU time, making pulse-fitting the next logical target for speed. My personal opinion is that pulse-fitting in java will always be too slow. Switching to something like multi-threaded C++ for this step is probably the best option.

mholtrop:
More information needed on the effects of this on the tracking. Especially in concert with Miriam's changes to the track fit.

cbravo135:
It is unclear to me how an argument against doing a calculation can be based on the percentage of the total processing time it takes. No matter what we run, everything will always add up to 100%. Should we not run the reco code at all since it takes up 100% of the reco CPU time? These numbers seem to indicate to me that the total time will only increase by 15% and I see no reasoning as to why a 15% increase in the total running time is actually detrimental.

PR #377, opened 2018-12-13
Added beam parameters obtained by Norman

rafopar:


mholtrop:
These changes are put in ISS370 by Norman.

PR #380, opened 2018-12-20
Iss361

rafopar:
==> Matt Solt fixed xhi2/NDF cut for 6 hit tracks.
==> In the steering file chi2 cut for 6 hits was changed from 72 to 84


rafopar:
There is conflict merging, I will cancel, and do pull request later

normangraf:
I would create a new git issue just for this one bug fix instead of reusing iss361.

PR #381, opened 2018-12-20
Iss361

rafopar:


rafopar:
Norman creating a new iss is the correct way, though the fix was trivial.

If you want new iss, then probably Matt Solt should do, as he knows where to

rafopar:
It is just one line, I will make an iss and do pull request

PR #383, opened 2018-12-20
Iss382

rafopar:
I think it is ready to be merged.

 I tested locally!


normangraf:
Because this changes the track cuts, the integration tests are failing. I am looking over the new histograms to make sure they are OK, before updating the reference sets.

normangraf (approved):
Checked and integration test reference histograms updated.

PR #388, opened 2019-01-11
Iss387

normangraf:
I have hard-coded the run-specific vertex positions as determined from V0 candidates by Matt Solt. I have enabled the use of the run-dependent X-Y beamspot and also updated the global 2016 target z position to -4.3mm in the steering file.


mholtrop:
Running a final test now. 👍 
There are also formatting changes in this pull. One set looked just weird, so I undid those :-).

mholtrop:
Passes most tests. Seems to do slightly better than before on some, which ironically makes it fail the tests:
Failed tests: 
  PhysRun2016MollerReconTest.testIt:66->comparePlots:93 expected:<2584> but was:<2595>
  PhysRun2016V0ReconTest.testIt:66->comparePlots:93 expected:<4792> but was:<4797>

mholtrop (approved):
Code is OK.

PR #395, opened 2019-01-28
fix ecal cluster energy correction for old, single beamgap geometry parameter

### LINKED ISSUE #394: Fix cluster correction for old, single "beamgap" parameter.
(Although might be best to completely remove support for single "beamgap" parameter)

baltzell:
Resolves #394 

Only affects geometries that use the old, single `beamgap` parameter and not the ones that use the separate `beamgapTop` and `beamgapBottom`.


normangraf (approved):
Should probably deprecate this at some point, but this fix should work for now.

PR #403, opened 2019-02-02
Always write out all vertex constrained particles.

mholtrop:
Basically, if you write one, you should write all of them, so any cuts should be done outside the loop. Currently the checks for chi-square and MaxVertexP are done on the UNCONSTRAINED collection.

### DIFF
--- a/recon/src/main/java/org/hps/recon/particle/HpsReconParticleDriver.java
+++ b/recon/src/main/java/org/hps/recon/particle/HpsReconParticleDriver.java
@@ -639,18 +639,16 @@ public class HpsReconParticleDriver extends ReconParticleDriver {
 
         // Create candidate particles for the other two constraints.
         for (Constraint constraint : Constraint.values()) {
-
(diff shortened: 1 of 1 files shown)


mdiamon:
I think this code change will cause fitVertex and makeReconstructedParticle to be called twice for Unconstrained (once before the loop and once in the loop)

mholtrop:
Miriam is correct, as it stands, it calls fitVertex and makeReconstructedParticle twice for the UNCONSTRAINED collection. I agree this is BAD.

Will fix this. In the meantime, some data was processed for Matt S to look at to make sure this change works for him.

mholtrop:
New version should address Miriam's observation that fitVertex and makeReconstructedParticle were called twice for the unconstrained vertex collection.

mdiamon:
I don't think this new version actually solves the calling-twice problem.

mdiamon (approved):

mholtrop:
Thank you Miriam, and thank you for your close inspection of the code.

PR #408, opened 2019-02-14
Iss320: Support offline conditions db usage (revised)

JeremyMcCormick:
This PR implements support for offline usage of the conditions database using SQLite.

The configuration of the connection parameters has been changed to only accept command line parameters, and support for providing these parameters via a text resource or file has been removed to simplify the db setup.  _**This change will require some updates to scripts that write to the database using the non-default account**_, but it should have no effect for the typical user connecting via the default read-only account.

The url, username and password can be provided as follows (this just replicates the default setup):

```
java -Dorg.hps.conditions.url=jdbc:mysql://hpsdb.jlab.org:3306/hps_conditions \ 
    -Dorg.hps.conditions.user=hpsuser \
    -Dorg.hps.conditions.password=darkphoton [...]
```

I checked that the SQLite db works by running this command:

```
java -Dorg.hps.conditions.url=jdbc:sqlite:hps_conditions.db \
-cp hps-distribution-bin.jar org.hps.evio.EvioToLcio \
-r -x /org/hps/steering/recon/PhysicsRun2016FullRecon.lcsim -d HPS-PhysicsRun2016-v5-3-fieldmap_v4_globalAlign -DoutputFile=scratch.slcio -n 10 \
events.evio
```

I also verified that the recon tests run using the JLab connection and that the same command line runs fine using the MySQL db.


mholtrop (approved):
Seems to work as described.
See: https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=245697776
For Doc on how to use.

PR #409, opened 2019-02-19
Iss 374

bloodyyugo:
Mostly, and most importantly, this changes the SVT timing corrections applied in the MC recon steering file.  Before this fix, we were seeing 2 peaks in the t0 for SVT hits separated by 24ns (somehow) due to the phase-adjustment correction which is not simulated in MC (so shouldn't be corrected for).  This 24-ns, two-peak structure isn't seen in the 2015 MC...though possible it's run condition (i.e. the value of the phase correction) dependent; I didn't try running the recon with different run assumptions (2015 I used the usual 5772, for 2016 I used 7796.  Anyway, the new steering file corrects just for the trigger timestamp

Here's what I need reviewers to do: 
Bradley -- make sure this builds & recos MC in a reasonable way e.g. how many reco SVT hit/tracks per event before and after change.  It would be great if you could do this for both some 2015 and 2016 data; I know what these changes should be so we can compare.  No new readout needed, just recon!

Norman & Rafo -- could you guys check the MC readout and recon files and see if they make sense compared to the data recon we are doing?  And that the readout constants put in for MC readout make sense?  Maybe this should be another issue...

### DIFF
--- a/analysis/src/main/java/org/hps/analysis/tuple/FittedSVTHitsTupleDriver.java
+++ b/analysis/src/main/java/org/hps/analysis/tuple/FittedSVTHitsTupleDriver.java
@@ -1,0 +1,250 @@
+package org.hps.analysis.tuple;
+
+import hep.physics.vec.BasicHep3Vector;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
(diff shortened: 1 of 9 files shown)


brigidcrowe (approved):
The changes so far see a track increase of >30%, with much better data agreement. The mean of the SVT time is still slightly shifted from data (~2ns), but it was discussed that this would be resolved in a separate new issue. 
[iss374Review.pdf](https://github.com/JeffersonLab/hps-java/files/2953634/iss374Review.pdf)

PR #412, opened 2019-03-18
iss338: Change all uses of DatabaseConditionsManager ctor to use getInstance() method

JeremyMcCormick:
The DatabaseConditionsManager constructor was made private, and public access is now achieved using the method `DatabaseConditionsManager.getInstance()` instead.  There is a single `INSTANCE` variable in the DatabaseConditionsManager class which will be instantiated if it does not exist.  The same conditions manager will then be used for the lifetime of the job/process.

I have checked that running the following tools seem to function properly with this change using OpenJDK:

- Evio2Lcio to output raw LCIO from EVIO
- JobManager to produce recon events from raw LCIO
- Conditions command line interface `print` command
- Detector converter from compact to LCDD (Sun JRE 1.8)

For detector converter, I used the class `org.lcsim.geometry.compact.converter.Main` as the HPS DetectorConverter seems to be broken, but it does not look like an issue having to do with the DatabaseConditionsManager.

Additional independent testing by a reviewer would be a good idea before merging this.

### DIFF
--- a/analysis/src/test/java/org/hps/analysis/MC/TrackExtrapTest.java
+++ b/analysis/src/test/java/org/hps/analysis/MC/TrackExtrapTest.java
@@ -51,7 +51,7 @@ public class TrackExtrapTest extends TestCase {
         String aidaOutput = "target/test-output/TrackExtrapPlots_" + fileName.replaceAll("slcio", "aida");
         LCSimLoop loop2 = new LCSimLoop();
         
-        final DatabaseConditionsManager manager = new DatabaseConditionsManager();
+        final DatabaseConditionsManager manager = DatabaseConditionsManager.getInstance();
         manager.addConditionsListener(new SvtDetectorSetup());
 
         loop2.setLCIORecordSource(inputFile);
--- a/analysis/src/test/java/org/hps/analysis/TupleDriverTest.java
+++ b/analysis/src/test/java/org/hps/analysis/TupleDriverTest.java
@@ -41,7 +41,7 @@ abstract public class TupleDriverTest extends TestCase {
         LCSimLoop loop = new LCSimLoop();
         loop.setLCIORecordSource(lcioInputFile);
         
-        final DatabaseConditionsManager manager = new DatabaseConditionsManager();
+        final DatabaseConditionsManager manager = DatabaseConditionsManager.getInstance();
         manager.addConditionsListener(new SvtDetectorSetup());
         
         loop.add(new org.lcsim.recon.tracking.digitization.sisim.config.ReadoutCleanupDriver());
--- a/conditions/src/main/java/org/hps/conditions/cli/CommandLineTool.java
+++ b/conditions/src/main/java/org/hps/conditions/cli/CommandLineTool.java
@@ -170,7 +170,7 @@ public final class CommandLineTool {
         LOGGER.info("Setting up the conditions manager ...");
 
         // Create new manager.
-        this.conditionsManager = new DatabaseConditionsManager();
+        this.conditionsManager = DatabaseConditionsManager.getInstance();
 
         // User specified tag of conditions records.
         if (commandLine.hasOption("t")) {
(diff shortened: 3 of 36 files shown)


mholtrop:
This issue causes a fail in the integration tests. Not sure about the details.
```[INFO] plugin ............................................. SUCCESS [  0.696 s]
[INFO] integration-tests .................................. FAILURE [26:41 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  29:57 min
[INFO] Finished at: 2019-03-18T12:03:13-04:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-failsafe-plugin:2.19.1:verify (default) on project hps-integration-tests: There are test failures.
[ERROR] 
[ERROR] Please refer to /data/HPS/hps-java/integration-tests/target/failsafe-reports for the individual test results.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :hps-integration-tests
```

JeremyMcCormick:
Okay, I will look into it...

JeremyMcCormick:
@mholtrop The failing test is a can of worms, unfortunately.  Its probably part of the reason I changed the initial class setup of DatabaseConditionsManager to have a public constructor, from what I remember.  The problem is with ClustererTest, which is using multiple LCSim loops to run different clustering configurations.  I don't know how to easily solve the crash due to the strange way that the conditions manager is activated in lcsim, which is part of the initial design and not my architectural choice.

Basically, the `setDetector()` method is activated within the lcsim event object *for every event*.  In a better architecture, it would be activated via a run header or other job management once for the entire processing with a set of conditions and activated only once.  What actually happens is that there is internal logic in the conditions manager to determine whether or not the run number and/or detector have changed, and only then are the conditions are reinitialized.  This activation includes the Driver classes being updated with new conditions in their `detectorChanged()` methods.  So what's happening here is that the conditions manager has already been initialized, and the clustering tests are using the same run number and detector.  So the re-initialization is not occurring for loops after the first one.

I'm not sure what to do about this, but the solutions could include:

- Excluding this test for now (don't like it as verifying working clustering is important).
- Adding a method to the conditions manager to create a new static instance method like `reset()` 
- Making the DatabaseConditionsManager constructor public but only using it in special cases such as this
- Making a public method like `setIsInitialized()` to toggle whether initialization should occur the next time `setDetector()` is called

Anything else I can think of would require some major framework changes to lcsim, and I do not think it is a good idea.

Please let me know which solution you think is best.  I'm leaning towards adding a method to toggle the initialization switch so that the conditions activation will fire the next time `setDetector()` is called even if the run number and detector is the same.

JeremyMcCormick:
@mholtrop I have fixed the above-mentioned problem by adding a static `reset()` method to the conditions manager which resets the current instance.  This can be called after initializing a new LCSim loop to force re-initialization of the conditions system.  It seems to fix the problems with the ClustererTest.  Please pull from the current iss338 branch, recompile hps-java from the base dir, and rerun the ClustererTest to confirm e.g.

```
cd hps-java; git checkout iss338; git pull
mvn clean install -DskipTests
cd integration-tests; mvn verify -Dit.test=ClustererTest
```

This seems like the most reasonable way to fix this issue without making changes to lcsim.

baltzell:
I confirmed ClustererTest passes.

baltzell (approved):

mholtrop (approved):
Seems all OK now.

PR #418, opened 2019-03-25
Iss166

mholtrop:
Big merge, but this is quite thoroughly checked and "master" was very recently merged in.
This branch passes all the integration tests.


brigidcrowe (approved):
Since thoroughly checked on its own, it was decided that this should be merged and any further testing/debugging done on the new master.

mccaky (approved):
This has been thoroughly checked at the event-by-event level by me. Matt Graham checked the SVT output and was satisfied, and the calorimeter output looked good to me.

There are very subtle differences with pile-up because the new readout outputs at a different time than the old one, but these appear to just be tails and not produce any significant effect on the final product. Output energy distributions for the calorimeter are extremely similar, and variance appeared to be on the order of eV, which is below the probable accuracy of our simulation anyway.

As such, this is likely good, and given that it is necessary for the hodoscope it is important to merge in and put under more full testing by the group. This is also important to allow branches to built off the new readout to support better truth propagation for the calorimeter and the hodoscope framework in a more formal, and less hacky, fashion.

Additionally, the original readout functionality was preserved, unchanged, so if an issue occurs, we can fall back on that without reverting the merge. The drivers have been moved to /readout/legacy/ in the steering-files package if needed.

As such, I support its merger into the master.

baltzell (approved):

PR #423, opened 2019-03-31
change tsCorrectionScale from 239ns to 171ns for new-style readout; l…

bloodyyugo:
…egacy the of ReconMC that works with old-style readout


brigidcrowe (approved):
Everything should now be correct in the PhysicsRun2016ReconMC steering files.

bloodyyugo:
So this includes the change in tsCorrectionScale then?


On Apr 1, 2019, at 10:48 AM, byale-jlab <notifications@github.com<mailto:notifications@github.com>> wrote:


@byale-jlab approved this pull request.

Everything should now be correct in the PhysicsRun2016ReconMC steering files.

mccaky (approved):
I independently verified that the timestamp difference is now 171 ns. Brad's changes are just reverting to the pre-merge state after it was overwritten, so that should be solid. Approved.

PR #444, opened 2019-04-15
Iss439: Move test to integration-tests and put output in correct dir.

JeremyMcCormick:
Minor change so that AIDA file generated by test goes in correct output dir.


mholtrop (approved):
Passes the tests. 
Not clear why this was needed, but the changes seem minor enough.

JeremyMcCormick:
> Passes the tests.
> Not clear why this was needed, but the changes seem minor enough.

We don't dump test outputs into the base module directories but into sub-directories of `target/` instead.

There's some test util classes I wanted to use to achieve this, which is why I moved this class from `analysis` into `integration-tests`.

Hope it is more clear now.

normangraf (approved):

PR #447, opened 2019-04-16
Create CrystalBallFunction.java

normangraf:
This is the simple one-sided Crystal Ball function. There's a main function that demonstrates its use.

### DIFF
--- a/util/src/main/java/org/hps/util/CrystalBallFunction.java
+++ b/util/src/main/java/org/hps/util/CrystalBallFunction.java
@@ -1,0 +1,269 @@
+package org.hps.util;
+
+import hep.aida.IAnalysisFactory;
+import hep.aida.IAnnotation;
+import hep.aida.IFitFactory;
+import hep.aida.IFitResult;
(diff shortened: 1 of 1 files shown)


bloodyyugo (approved):

util/src/main/java/org/hps/util/CrystalBallFunction.java
```diff
@@ -0,0 +1,269 @@
...
+    @Override
+    public void setTitle(String string) throws IllegalArgumentException {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    @Override
+    public double value(double[] x) {
+
+        double s = (x[0] - xmean) / sigma;
+        double A = pow((n / abs(alpha)), n) * exp(-0.5 * abs(alpha) * abs(alpha));
```
omar-moreno, line 53: Might be worth checking if recursive power function is faster. Might be beneficial when doing multiple fits.

omar-moreno (approved):

PR #449, opened 2019-04-23
The new detector 'HPS-PhysicsRun2019-v1-4pt5' for 2019 run is added, …

rafopar:
…which is different from the previous version (HPS-PhysicsRun2019-v0-4pt5) by a shift of L1 Hodo tiles by 2 mm towards positron side


mholtrop (approved):
Yay.

normangraf (approved):
Should probably leave this issue open for a while in case there are additional changes needed for the SVT. But this makes sense to merge in now.

PR #459, opened 2019-05-06
Fix crash for low numbered runs where hodoscope is not initialized.

mholtrop:
Very simple bug fix.

### DIFF
--- a/conditions/src/main/java/org/hps/conditions/database/DatabaseConditionsManager.java
+++ b/conditions/src/main/java/org/hps/conditions/database/DatabaseConditionsManager.java
@@ -745,7 +745,7 @@ public final class DatabaseConditionsManager extends ConditionsManagerImplementa
         }
         this.registerConditionsConverter(this.svtConverter);
         this.registerConditionsConverter(this.ecalConverter);
-        this.registerConditionsConverter(this.hodoscopeConverter);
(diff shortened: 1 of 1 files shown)


JeremyMcCormick (commented):
This isn't necessary, and I don't think this PR should be approved.  The proper way to fix is just using a valid run number which is not from the Test Run.

mholtrop:
Not fixing this means that the code will not work for _any_ test run data, i.e. all runs with low run numbers. That is OK, in my opinion, but in that case, _all_ test run clutter should simply be ripped out off the code base instead of cluttering up everything. There is no need to keep code in the code base that is broken and we are not willing to accept simple fixes to.

JeremyMcCormick:
Oh, I see now there's a bug, because it's registering a null pointer, so
this fix makes sense.  Thanks for
catching this.

I'm for ripping out Test Run code and tables from
the conditions system as the chance anything works with it now in the
framework is zero, and no one has looked at that data in years.  But now is
probably not the time.

JeremyMcCormick (approved):

mholtrop:
Hi Jeremy,

Yes, I agree, and I agree this is not the time to decide this, but we can at least not worry about support. We’ll decide officially after the run :-).
We have all he old code in old releases, and a backup of the DB, so if ever there is a really good reason to go back to that data, it will still be possible. 

Best,
	Maurik

normangraf (approved):

PR #463, opened 2019-05-07
Iss411: Fix detector converter for recent JDK

JeremyMcCormick:


mholtrop (approved):
Verified with Oracle Java JDK 12.

normangraf (approved):

PR #467, opened 2019-05-10
Iss434: Add EVIO Data Converter for Hodoscope

rafopar:
@mholtrop Please select the **squash** option when merging.

**This PR should not be merged until the main test passes.**  (See below for more details.)

---

This PR has updates for converting Hodoscope raw data to LCIO.

Some important things to note with this PR:

- This should be merged after PR #471.

- There may be some conflicts between shared files on these two branches that need to be resolved, which I can help with.

- Bank tags in the converter are currently set at values for "alcove data" and so the test case fails.  This should be resolved before merging by either updating the test data file or reverting this change and patching the bank tags later in a separate PR.


JeremyMcCormick:
Some important things to note:

This should be merged after PR #471.

There may be some conflicts between the changes in these two branches that need to be resolved, which I can help with. 

@mholtrop Please select the **squash** option when merging.

JeremyMcCormick:
@rafopar 

Your latest set of commits breaks the test completely as no readout hits are produced from the Hodoscope only test file you provided me.

You can run as follows:

```
cd integration-tests; mvn verify -Dit.test=HodoscopeDataConverterTest
```

When I add a check to the Driver to make sure there are readout hits, I see that the collection is always empty now.

I assume this is because you have changed the bank numbers to match real data.  We need to update the test EVIO file so that it will work with these new tags, or we should revert the code until we actually have data with this new structure.  Reviewers need to be able to run the test in order to verify that your changes are correct.

JeremyMcCormick:
The issue reported by @normangraf where EVIO data files without Hodoscope data could not be read should be fixed by this commit:

https://github.com/JeffersonLab/hps-java/pull/467/commits/7ab8b3394dfd0dc14619a9d82751576571d8e1f0

I have added a check on purpose to the test case which throws an error if there are no readout hits in the event.  We need to update the test data to fix this.

JeremyMcCormick:
I resolved a conflict with a detector data file by doing:

```
git checkout master -- detector-data/detectors/HPS-PhysicsRun2019-v1-4pt5/HPS-PhysicsRun2019-v1-4pt5.lcdd
```

This overwrites whatever was on the branch for this file with the copy from master.

I hope this is what we want - we can fix-up later if this causes issues (this detector is under development anyways and not currently usable as far as I'm aware).

normangraf:
You're correct @JeremyMcCormick, that that detector is a work in progress. It is usable for simulation, but the current SVT reconstruction does not yet work with the "slim Layer1".

normangraf:
I just updated the test file for the HodoscopeDataConverterTest, so that now completes successfully. However, the 2015 and 2016 integration tests throw errors of the type:
[SEVERE] [org.hps.evio] Error making Hodo hits.
java.lang.NullPointerException
        at org.hps.evio.HodoEvioReader.daqToGeometryId(HodoEvioReader.java:247)
        at org.hps.evio.HodoEvioReader.makeWindowHits(HodoEvioReader.java:226)
        at org.hps.evio.HodoEvioReader.makeHits(HodoEvioReader.java:128)
        at org.hps.evio.LCSimEngRunEventBuilder.makeLCSimEvent(LCSimEngRunEventBuilder.java:174)
        at org.hps.evio.EvioToLcio.run(EvioToLcio.java:608)
        at org.hps.evio.EvioToLcio.main(EvioToLcio.java:92)
        at org.hps.test.it.EngRun2015V0ReconTest.testIt(EngRun2015V0ReconTest.java:48)

Is this now already fixed in master?

normangraf:
All three of the hodoscope tests (HodoscopeDataConverterTest, HodoscopeSimDataTest and HodoscopeReadoutTest)  run to completion.

Please reduce the verbosity of the HodoscopeReadoutTest. Is it also necessary to run over 1140300 events for this test? 

I see that there are conflicts with evio/src/main/java/org/hps/evio/LCSimEngRunEventBuilder.java . I believe resolving that will also resolve the issues with the 2015 and 2016 integration tests.

normangraf (approved):
See my latest comments regarding the test and the one conflict with evio/src/main/java/org/hps/evio/LCSimEngRunEventBuilder.java 

Other than that I think we should proceed.

rafopar:
Hi @normangraf , which are conflicting liens in the `evio/src/main/java/org/hps/evio/LCSimEngRunEventBuilder.java` file?

rafopar:
Ah ok @normangraf nevermind, I saw it.

mholtrop (approved):
Code was cleaned up some, and now passes all test. 
A new issue will deal with further improving the EVIO reading program flow.

PR #475, opened 2019-05-14
Minimal fixes to ensure that the ECAL cosmics calibration code works with the 2019 setup

andrea-celentano:


baltzell (approved):

JeremyMcCormick (dismissed):
I would request one change here.

You have copy and pasted many checks like this throughout the code:

```
if (ihit == null) {                  
    pulse[ix][iy] = 0;
    signal[ix][iy] = 0;
    continue;
}
```

Instead, can you just initialize the pulse and signal arrays to all zeroes to start with?

Then I believe it should be unnecessary.

mholtrop (approved):
I disagree with Jeremy's comment. Even if the arrays are initialized, if the hit that the code attempts to get from LCIO is Null, you would still get a crash, no matter how much initialization was done.
The code is good as is.

PR #481, opened 2019-05-16
Hodoscope Readout Chain Fixes

mccaky:
A quick set of fixes for the recent iss420 merge. These include:

- `ReadoutDataManager` now uses the `Logger` class to handle print statements and is much quieter by default.
- The outdated steering file `Hodoscope_2pt3_EcalOnlyTrigger` has been removed.
- The units for the hodoscope plots in `HodoscopePlotsDriver` is now MeV.
- Readout raw converter implementations now require that NSA and NSB be defined in the steering file. It is necessary that both the digitization driver and raw converter driver use the same values, and it is easy to miss that this is the case. This makes it harder for that to happen.
- NSA and NSB for the hodoscope raw converter are now set to the correct values, fixing an issue where some hodoscope converted hits had negative energy.
- All readout steering files have been updated to meet the new raw converter requirements.
- Readout steering files had all tabs replaced with spaces.

Note that the readout raw converters are not used for recon. These should be switched over at some point, but at the present, these changes do not affect recon at all.


mccaky:
Note that there is a known issue with the preprocessing driver where hit energies and times are gaining minor shifts in their values during conversion to `SimCalorimeterHit`. This appears to be an issue with the underlying behavior of the LCSim object and not related to the actual preprocessing driver. This is being handled separately in iss479.

JeremyMcCormick:
https://github.com/JeffersonLab/hps-java/pull/481/commits/680757916ab03db262e414286468da1efecacb9a

👍 🥇

JeremyMcCormick (commented):
These changes look fine.  The readout test passes and plots look okay.

JeremyMcCormick (approved):

rafopar (approved):
They look good to me too

PR #489, opened 2019-05-22
Fix Steering File Issues

mccaky:
A few changes to drivers were not propagated into the steering files during the iss420 merge. This corrects it. All changes are to the readout steering files only.


mholtrop (approved):
Tested each on a spaced SLIC output file. They seemed to work fine now.

brigidcrowe (approved):
I checked the iss487 readout and it works now, so this should be merged ASAP.

PR #492, opened 2019-05-27
Hodoscope Readout Minor Fixes

mccaky:
Fixed some minor issues:
- The collections printer utility would fail if a collection that was marked for printing did not exist in a given event. This is changed so that it now skips that event entirely.
- The collections printer utility would incorrectly mark the particle ID and time for an `MCParticle` as having units of GeV. This has been corrected - particle ID is now unitless and time is now in units of ns.
- The digitization driver would fail if truth information was persisted and an MCParticle or one of progenitor particles had multiple parents. The driver now correctly handles particles with multiple parents.

### DIFF
--- a/analysis/src/main/java/org/hps/analysis/util/PrintCollectionsUtility.java
+++ b/analysis/src/main/java/org/hps/analysis/util/PrintCollectionsUtility.java
@@ -288,8 +288,8 @@ public class PrintCollectionsUtility {
         StringBuffer outputBuffer = new StringBuffer();
         String indent = getIndent(indentLevel);
         
-        outputBuffer.append(String.format(indent + "Type     :: " + yellow("%d") + " GeV%n", particle.getPDGID()));
-        outputBuffer.append(String.format(indent + "Time     :: " + yellow("%f") + " GeV%n", particle.getProductionTime()));
+        outputBuffer.append(String.format(indent + "Type     :: " + yellow("%d") + "%n", particle.getPDGID()));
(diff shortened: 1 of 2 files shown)


mccaky:
This makes very minor changes - I have verified that the print collections utility now behaves correctly, but if someone wants to run it, it can be engaged via:

`java -cp $HPS_JAVA org.hps.analysis.util.PrintCollectionsUtility -d $DETECTOR_NAME -R $RUN_NUMBER -i $INPUT_FILE -p $COLLECTION_NAME`

where `-d` specifies the name of the detector associated with the input file, `-R` specifies the run number associated with the input file, `-i` specifies the input file, and `-p` (which is optional) specifies which collections to print out. Note that only certain object types are supported. These are `CalorimeterHit`, `SimCalorimeterHit`, `GenericObject`, and `MCParticle`. The error can be reproduced by trying to print a collection in a spaced MC file, as these do not contain collection in the empty spacing events. It will crash in the master version, and not in this version.

The digitization driver can be tested by running the steering file `SimpleHodoscopeTrigger` over a data file that contains `MCParticle` objects which have more than one parent. The recently produced A' data on the JLab tapes at `/mss/hallb/hps/production/2019Run/noTilt/slic/ap/4pt55/HPS-PhysicsRun2019-beta1-4pt5/` meets this criteria.

mholtrop (approved):
Minor fixed of Readout system, only affects when truth data is retained.

normangraf (approved):

PR #500, opened 2019-06-04
Iss496

rafopar:
The purpose of this branch is to make it to be able to read VTP banks from the evio file.

It reads VTP banks from both (bot and top) rocs, and stores them in the output scliio file as a genericCollection object.

This collection has only int values.
There are two collections which have same number of integer elemts, one of them contins rocIDs, and the other contains actual words.

The integration test didn't give any issue.

Conversion is tested from several VTP words, and seems conversion works as it is expected

### DIFF
--- a/evio/src/main/java/org/hps/evio/LCSimEngRunEventBuilder.java
+++ b/evio/src/main/java/org/hps/evio/LCSimEngRunEventBuilder.java
@@ -12,7 +12,7 @@ import org.hps.record.epics.EpicsEvioProcessor;
 import org.hps.record.evio.EvioEventUtilities;
 import org.hps.record.scalers.ScalerData;
 import org.hps.record.scalers.ScalersEvioProcessor;
-import org.hps.record.svt.SvtEvioExceptions.SvtEvioReaderException;
+//import org.hps.record.svt.SvtEvioExceptions.SvtEvioReaderException;
 import org.hps.record.triggerbank.AbstractIntData;
 import org.hps.record.triggerbank.AbstractIntData.IntBankDefinition;
 import org.hps.record.triggerbank.HeadBankData;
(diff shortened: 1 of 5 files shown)


JeremyMcCormick:
Is there a simple unit test showing how to get the values from the event?  

If not should probably add one.

Where is some sample data for testing?

rafopar:
I haven't made a testunit before. I don't exactly know how to do it.
Should it be a code in integration-tests/src/test/java/org/hps/test/it ?

Although, I want to mention that master itself was braking when I run over these files.
It was giving NullpointerException from SVT readout part, and therefore during development, I commented SVT readout part in the 
` svtReader.makeHits(evioEvent, lcsimEvent);`

then when the VTP readout is complete, I uncommented it back.

The exception that master gives is the following

Exception in thread "main" java.lang.NullPointerException
        at org.hps.evio.SvtEvioReader.setupDaqMap(SvtEvioReader.java:104)
        at org.hps.evio.AbstractSvtEvioReader.makeHits(AbstractSvtEvioReader.java:165)
        at org.hps.evio.LCSimEngRunEventBuilder.makeLCSimEvent(LCSimEngRunEventBuilder.java:189)
        at org.hps.evio.EvioToLcio.run(EvioToLcio.java:608)
        at org.hps.evio.EvioToLcio.main(EvioToLcio.java:92)

rafopar:
Hi @JeremyMcCormick , I put the input evio file, the output test file, and the jar file, in the following directory

/work/hallb/hps/rafopar/tests/iss496

And below is the command that I used to run

`java -cp hps-distribution-4.4-SNAPSHOT-bin.jar org.hps.evio.EvioToLcio -d HPS-PhysicsRun2016-Pass2 -R 8099 hps_008099_tagged.evio.0 -l testout.slcio -n 2500 -e 500`

using run numbers below 9000 doesn't crash SVT, so you can use e.g. 8099

JeremyMcCormick (approved):
Code is a little rough, especially some of the formatting, but I guess we won't worry about it.

I'm able to run the example command line, write a raw LCIO file, and then read it into JAS3 and see the VTP bank data (no idea if it is correct or not but I'll trust Rafo on this as I've no good way to check).

All tests seemed to pass when I ran the dev build in Jenkins.

Approved!

normangraf (approved):
Works as advertised. No adverse effects. Approved.

PR #503, opened 2019-06-05
Instead of one ID, two IDs are calculate.

rafopar:
// Unlike to the ECal case, where each detector element is readout with a single channel
        // here for the hodoscope, there are detector elements (tiles) that are readout with two different channels
        // 
        // The code previously was assigning the hole value to 0, 
        // That way you would get an iD, which is ok for the detector element identification, however
        // This id is not correct, when one wants to know which PMT channel is actually this hit belongs to.
        // In particular during the recon, I was getting an error, that it can not find a cellID

        // Now as a solution (Possibly not the best solution), this method will return two IDs
        // The 1st one, is calculated taking into account the ix,iy,layer and the hole,
       while the 2nd one is calculated with the "hole=0" assumption.

        // Further depending on the intention, one can use ids.get(0) if one wants to check the hit ID, or ids.get(1), if one
        // wants to identify the detector element.

### DIFF
--- a/evio/src/main/java/org/hps/evio/HodoEvioReader.java
+++ b/evio/src/main/java/org/hps/evio/HodoEvioReader.java
@@ -33,9 +33,9 @@ import org.lcsim.lcio.LCIOConstants;
 
 /**
  *
- * This class is similar to the ECalEvioReader.  It was copied and modified
- * to work with Hodoscope data.
- * 
+ * This class is similar to the ECalEvioReader. It was copied and modified to
+ * work with Hodoscope data.
+ *
  * @author rafopar
  */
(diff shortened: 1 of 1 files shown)


mholtrop:
The code looks OK, but currently I cannot test this. (Not until Friday)
If this is in a hurry, please ask other reviewer.

rafopar:
Thank you @mholtrop Maurik, yes I understand you are extremely busy with the MENU 2019,
If Jeremy also will not find issue, then we perhaps can ask someone else to review it

normangraf (approved):
The code looks like it is doing what it is advertised to do. All the tests run, so it does not affect older data.

mholtrop (approved):

PR #516, opened 2019-06-20
VTP bank rocIDs are changed, the values before was for simulated data

rafopar:
A minimal change.

VTP rocIDs values were wrong, it was for simulated data.
these two numbers are updated.


normangraf (approved):
I have no way of independently verifying this, but I trust that these IDs are correct.

JeremyMcCormick (approved):
Assuming this has been tested.

rafopar:
Yes I tested, and it works, correctly report VTP banks

PR #531, opened 2019-07-02
Revert "SVT EVIO reader for the 2019 physics run."

JeremyMcCormick:
Reverts JeffersonLab/hps-java#520

This PR broke all the recon integration tests which use 2015 and 2016 data.


bloodyyugo (approved):
This change is in the Run2019Master branch and works for 2019 data; apparently causes a memory leak in 2016 data so that needs to be fixed before getting into the Master.  

For 2019 data, use Run2019Master

omar-moreno:
@JeremyMcCormick Can you point me to the specific test that is failing? I ran SvtEvioReader and I'm not seeing a memory leak or any errors.  Also, what file are you using to test?

JeremyMcCormick:
> @JeremyMcCormick Can you point me to the specific test that is failing? I ran SvtEvioReader and I'm not seeing a memory leak or any errors. Also, what file are you using to test?

@omar-moreno 

Run any of the recon ITs such as:

```
cd integration-tests; mvn verify -Dit.test=EngRun2015MollerReconTest
```

It will run but takes longer and longer to process each event.  I let it sit there for several hours until I killed the process.  Normally, *all* the tests run in < 1.5 hours, so there is something wrong here, probably a memory leak due to bad allocation of an array (but I don't know exactly).

If I revert to a version of the master just before your changes were merged, I do not see this problem with the tests:

```
git checkout ea7b71b75df1395f2ed5eb5a66e173223b1e9893
```

This indicates that the updates to the SVT reader caused this issue.

mholtrop (approved):

PR #533, opened 2019-07-04
Remove useless print outs that were annoying me.

JeremyMcCormick:


mholtrop (approved):
Don't know what _those_ were so annoying, but hey, change looks fine.

baltzell (approved):
definitely pointless since direct to stdout without any info

PR #547, opened 2019-07-15
Iss544

mholtrop:
Allow the code to carry on when a bad MakeGblTrack call returns a null pointer. Simple fix.
Merges into Run2019Master.

Norman will be looking into the details of why that null pointer is there is the first place, but it is likely from some invalid or poor seed track, so nothing to worry about.


normangraf (approved):
Looks fine. It's infrequent now, but we need to keep an eye on the recon log files to see how often this happens. I have a feeling it will go away with different track-finding strategies and tighter cuts.

PR #548, opened 2019-07-15
Handle NaN exception thrown by HelixFitter

mholtrop:
After iss544, this can skips the next issues: NaN in the helix parameters for specific seed.

The problem was briefly discussed in the tracking meeting. It is probably not worth actually fixing this issue, since there are most likely deeper issues with the seed that caused the helix fit to fail so that it adds NaN for the parameters.


normangraf (approved):
Looks fine. It's infrequent now, but we need to keep an eye on the recon log files to see how often this happens. I have a feeling it will go away with different track-finding strategies and tighter cuts.

PR #592, opened 2019-08-17
iss588: Parse 2019 DAQ configuration banks

tongtongcao:
Modify TriggerConfigEvioReader.java to read 2019 DAQ configuration evio data and develop a parser to parse 2019 DAQ configuration banks

### DIFF
--- a/evio/src/main/java/org/hps/evio/TriggerConfigEvioReader.java
+++ b/evio/src/main/java/org/hps/evio/TriggerConfigEvioReader.java
@@ -9,6 +9,7 @@ import org.hps.record.daqconfig.EvioDAQParser;
 import org.jlab.coda.jevio.BaseStructure;
 import org.jlab.coda.jevio.EvioEvent;
 import org.lcsim.event.EventHeader;
+import org.hps.record.daqconfig.EvioDAQParser2019;
 
 /*
  * Search for a configuration bank in EvioEvent, and, if found, create an instance of
@@ -25,6 +26,10 @@ import org.lcsim.event.EventHeader;
  * 
  *  
  * @author <baltzell@jlab.org>
+ * 
+ *  * <code>EvioDAQParser2019</code> is parser for 2019 trigger configuration banks.
+ *  
+ * @author Tongtong Cao <caot@jlab.org>
  */
 public class TriggerConfigEvioReader {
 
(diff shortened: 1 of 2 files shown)


baltzell (approved):

evio/src/main/java/org/hps/evio/TriggerConfigEvioReader.java
```diff
@@ -39,28 +44,54 @@ public void getDAQConfig(EvioEvent evioEvent, EventHeader lcsimEvent) {
...
-                continue;
-            int crate = bank.getHeader().getTag();
-            for (BaseStructure subBank : bank.getChildrenList()) {
-                if (subBank.getHeader().getTag() == EvioDAQParser.BANK_TAG) {
-                    if (subBank.getStringData() == null) {
-                        Logger.getLogger(this.getClass().getName()).log(Level.SEVERE,
-                                "JEVIO can't parse DAQ Config bank.  Event Number "
-                                        + evioEvent.getEventNumber());
-                        return;
+        else if (lcsimEvent.getRunNumber() >= 4043 && lcsimEvent.getRunNumber() < 9000) {
```
JeremyMcCormick, line 47: I'd suggest you don't hard-code checks on run numbers like this.  Instead, update this class with a flag or other configuration based on the conditions system hooks.

tongtongcao: Do you mean that I set a flag according to run number ranges, and then put expression with flag into "if condition"?
What kind of condition parameters/hooks can be used to identify run number ranges?

tongtongcao: Ah... I think I got what you said. You mean use like DatabaseConditionsManager.isPhys2019Run(lcsimEvent.getRunNumber())
instead of 9000

JeremyMcCormick: Whatever class is creating this one (event builder?) should have a `detectorChanged` or other conditions system hook where you can set some configuration on the parser.  I would either set a boolean flag or better yet, just have a small class that implements the correct behavior, e.g. register a specific object depending on what run number is seen in the conditions callback method.  We try not to hard-code specific run numbers within our algorithms.

If what you have works, then we can merge without doing this.  I guess for now it is okay if we just want to get something working.  (I see this has already been approved so I assume it works.)

JeremyMcCormick: Not exactly...

In `LCSimEngRunEventBuilder`, there is a method `conditionsChanged` where you should configure your class depending on what run number is seen there.  You can get the run number from the manager there.

This keeps all the conditions configuration/checking in one place and not scattered around in many places within the code.

In fact, shouldn't there be an entirely separate `TriggerConfigEvioReader` class for 2019? 

I don't think it is a great idea to modify the one that's there with a bunch of if/then statements depending on the run number.  It would be better to have entirely separate top-level classes.

tongtongcao: Agree. Could I merge this branch now since we need this parser for some studies of trigger validation? And then I will build a new branch to make changes according to your suggestions.

tongtongcao: The changes will not affect application of the parser.

PR #593, opened 2019-08-25
Hodo developement

rafopar:
Jeremy thanks for finding this and reporting.

I added that file. Since the previous pull request there was other changed in the HodoscopePlots.java  file.


JeremyMcCormick (approved):
compiles now

PR #606, opened 2019-09-19
Iss601

JeremyMcCormick:
This PR revises a number of classes involved with SVT EVIO data conversion so that they no longer cause severe memory leaks and hanging of data processing. 

Direct generation of exceptions was removed from header error checking in favor of using a more lightweight data class. 

A memory leak in the SVT EVIO reader where header data was not being cleared between events was found and corrected.

The recon ITs on 2015 and 2016 data work now from Run2019Master and no longer hang indefinitely. Processing time is about 14 events a second, and 4000 events from the 2016 run were processed in under 5 minutes, including initialization time.

### DIFF
--- a/evio/src/main/java/org/hps/evio/AugmentedSvtEvioReader.java
+++ b/evio/src/main/java/org/hps/evio/AugmentedSvtEvioReader.java
@@ -1,6 +1,3 @@
-/**
- * 
- */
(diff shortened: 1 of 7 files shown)


normangraf (approved):
Builds and all tests pass.

PR #620, opened 2019-10-01
Iss617

mrsolt:
Changing the strip length to 15*2 mm, and cleaning up some deprecated code.


bloodyyugo (approved):
Using DQM code, I see that the 1d clusters have moved out in x, so I guess is ok.  Tracks/V0s change very little

normangraf (approved):
Appears to do what it claims to do. I'm surprised it doesn't have a larger effect on the tracks. But will approve.

PR #623, opened 2019-10-14
Iss622

pbutti:
The millepedeLayer ID was off alignment for 2019 geometry because IDs 9 and 10 were skipped, while they should have been assigned to Layer 5. Added a check on the compact.xml in getMillepedeLayer in HPSTracker2014GeometryDefinition that tries to get the year. 
*Assumption*: The compact.xml specifies the year in the tracker subdetector type attribute, otherwise 2016 geometry is assumed.

### DIFF
--- a/detector-model/src/main/java/org/lcsim/geometry/compact/converter/HPSTracker2014GeometryDefinition.java
+++ b/detector-model/src/main/java/org/lcsim/geometry/compact/converter/HPSTracker2014GeometryDefinition.java
@@ -2200,7 +2200,7 @@ public class HPSTracker2014GeometryDefinition extends HPSTrackerGeometryDefiniti
                 layerOffset = 7;
             }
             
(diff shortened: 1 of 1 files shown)


pbutti:
Adding reconstruction log files to show the millepede IDs after the fix. 2016 correctly ends at 18, while 2019 ends at 20 (before was 22, skipping 9 and 10)
[log_MillepedeLayerCheck_2019.txt](https://github.com/JeffersonLab/hps-java/files/3726012/log_MillepedeLayerCheck_2019.txt)
[log_MillepedeLayerCheck_2016.txt](https://github.com/JeffersonLab/hps-java/files/3726013/log_MillepedeLayerCheck_2016.txt)

normangraf (approved):
Checking on the year is a bit of a hack, but the error in the logic that was fixed was real, so I'm OK with this.

bloodyyugo (approved):

PR #627, opened 2019-10-20
Iss626

mrsolt:
Fix duplicate variable in tuple maker (eleP and posP to eleStartP and posStartP).

Remove cut tuple in steering files.


bloodyyugo (approved):

normangraf (approved):
Good to go.

PR #631, opened 2019-10-23
Fix pulser trigger rate calculation

cbravo135:
Events in readout are 2 ns each, and not 1 ns each. Small change to fix this.


mholtrop (commented):
I think that the issue here of a factor 2 versus 4 appears to be one of *documentation* rather than code. I could well be misreading the comment, but my interpretation is:  the method itself talks about "clockCycles", which I assume to be FADC clock cycles. The FADC clock cycles are 4ns apart. The comments in the class header state that the code will trigger every nth event. "Events" is this case would be MC beam events, which are 2ns spaced.  
It then seems to me that the original 2 was indeed the correct factor, since this results in a n*2 events*2ns = n*4ns spacing. 
I found only one steering file that uses this (), which states:
```        <driver name="PulserTrigger" type="org.hps.readout.trigger.PulserReadoutDriver">
            <!-- Units of clock-cycles (4 ns, 2 events). -->
            <pulserRate>50000</pulserRate>
        </driver>
```
Which seems consistent with how I read the code.

If my argument above is not correct, could you please clarify this for me?

cbravo135:
"It then seems to me that the original 2 was indeed the correct factor"
The original factor is 4, this PR changes it to 2.

mholtrop (approved):
Correct!
I must have been looking at the wrong branch when checking the details in the code.
This does not need 2 reviewers.

rafopar (approved):

PR #635, opened 2019-11-01
Iss630

pbutti:
The changes in this merge request are for adding the Multiple Coulomb Scattering  computation for traversed sensors where hit-on-track is not present:
- Before MCS was only added for measurements on track, while now it's added also for holes-on-track (track hypothesis traverses a sensor but there isn't a hit associated). 
- This "new" behaviour is OFF by default (original behaviour is used) but can be turned on by setting includeMS flag in the GBLRefitterDriver.
- Checked that doesn't break things and the effect (see:
https://confluence.slac.stanford.edu/display/hpsg/10.28.2019+Weekly)
- Added control on GBL refit iterations from steering file
- Moved out GlobalDers from the refitter to stand-alone .java just to clean up code
- Modified GBLStripClusterData to hold information on millepedeID, volume and if it's a hit+scatter or scatter only

- Tested compilation and run-time

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/gbl/GBLRefitterDriver.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/gbl/GBLRefitterDriver.java
@@ -1,9 +1,7 @@
 package org.hps.recon.tracking.gbl;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 
 import org.apache.commons.math3.util.Pair;
 import org.hps.recon.tracking.MaterialSupervisor;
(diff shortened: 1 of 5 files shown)


normangraf:
Have you built and run all the tests (component as well as integrated) for this issue?

pbutti:
Hi Norman,

thanks for the comment. Currently running them.

pbutti:
Ran the tests. The results are available:
/nfs/slac/g/hps2/pbutti/hps-java/integration-tests/target/failsafe-reports/
They failed because of DiskQuotaExceeded . I've tried cleaning up the directory (removed 6GB) but didn't help. I'm running on cent7a @ SLAC. 
@normangraf  any suggestion on how to fix this issue? Perhaps I'm just running in the wrong place

mholtrop (approved):
Passes all tests.

normangraf (approved):
Looks good. Please proceed.

PR #642, opened 2019-12-03
Fix to un-initialized matrix in gbl when interpolating 

pbutti:
This fix will not affect track reconstruction (nOffset is >=0 in hps track reconstruction), however some derivative matrices are computed with matrices that are zero. This affects when checking the GBL internal math with the gbl examples. 
After this fix I am able to reproduce the results of C++ implementation of the gbl examples on svn desy repo. 
Fix should be trivial. Checked that the results are correct using gbl C++ example code.


mholtrop (approved):
Minimal change, seems fine. 
Tests are still running, but seems fine.

normangraf (approved):
Looks reasonable.
From your comment, it appears that you have coded up the C++ GBL tests in java. If so, please commit these tests as well.

pbutti:
Thanks guys for the review. 
I will push the example by today in a separate commit.

PR #646, opened 2019-12-07
Iss640

tongtongcao:
1) Some cute in the MC recon steering file PhysicsRun2016FullReconMC.lcsim are tighter, than the one for the pass4 data. We set these cuts to be at least the same as in data. Additionally, we fix an issue for track cluster time offset.
2) The new readout steering file PhysicsRun2016TrigPairs1_NoTruth.lcsim is produced by Kyle to exclude truth information
3) Set  <cutTuple>false</cutTuple> in the tuple steering file MakeTuplesMC.lcsim for tuple data production.


rafopar (approved):
This branch was tested, and produced MC readout and recon files have issues fixed.
Those issues include:
- Large event size
- Track time not being centered at 0
- Because of track time not being centered at 0, the cluster-track time difference cut in MOUSE, also cut a little bit tight the signal.

bloodyyugo (approved):
as long as Rafo says it's ok, it's ok with me

PR #648, opened 2019-12-17
Iss549

normangraf:
Provided default steering file for 2019 data processing.

Added a few switches to allow the ECal cluster corrections to be turned off for initial 2019 reconstruction (on by default in 2015/2016) in steering file.

Provided a few example analysis drivers.


bloodyyugo (approved):
I ran this branch with the analysis drivers turned on...looks ok to me.

normangraf:
Thanks Matt. I also ran it with the analysis drivers, but some of them were requesting to open windows, which isn't going to work when running in batch so I commented them out in the "official" steering file.

normangraf:
Please note that the 2019 integration test does not yet include any verification of the output. Hopefully things will be improving so fast that we would be annoyed by having to update the reference histograms so often...

baltzell (approved):

mholtrop (approved):

PR #649, opened 2019-12-19
Iss621

tongtongcao:
The package org.hps.record.daqconfig2019 includes all codes for processing 2019 DAQ configuration, including DAQ configuration parser, configuration manager, configuration driver, storages of parameter values, interface for accessing parameters, etc.
Introductions to classes in the package are listed in the item DAQConfigDriver2019 of the section "Driver lists in plan" at the confluence page https://confluence.slac.stanford.edu/display/hpsg/Trigger+Diagnostics+2019


mccaky:
We'll need to delete all of the .metadata files in the request before can merge the pull request. That's all local Eclipse setup stuff, and shouldn't be a part of the repository itself.

tongtongcao:
> We'll need to delete all of the .metadata files in the request before can merge the pull request. That's all local Eclipse setup stuff, and shouldn't be a part of the repository itself.

Thanks. 
.metadata files have been removed.

mccaky:
I'm still working on my full review, but here are a few quick comments.

**HodoEcalCoincidence2019**
What information is this class supposed to store? Presently, it does not actually seem to define either mutator or selector methods for any values. I assume that it is intended to store whether or not a calorimeter/hodoscope coincidence is needed for a trigger. If that is the case, it would be better to write it as an `AbstractConfig<Boolean>` with a single parameter that is either `true` or `false`.

**TSConfig2019**
I don't see any issues with the implementation itself. Everything looks good there! I think it would be useful, however, to include some public static final member variables to map what each index in the array corresponds to, so others who may work with the code can easily link the two. For example:
```
		public static final int SINGLE_0_TOP = 0;
		public static final int SINGLE_1_TOP = 1;
		...
```
Then, when you get the array, you can just reference a given value with:
```
		int[] prescales = tsConfig.getTSPrescales();
		int st0 = prescales[TSConfig2019.SINGLE_0_TOP];
```
This makes it so that the user doesn't need to directly reference the source code to know which array entry corresponds to what.

**PDECutConfig2019** / **FEETriggerConfig2019**
This isn't strictly necessary to change, since it is really a bit pedantic of me, but Java standards generally make public static final variables in all caps. So, for instance, `parC0` --> `PAR_C0`. I'll still pass the pull request if you ignore this one, though. :)

mccaky:
One of the bigger issues I see that we should address is that we want to minimize the repetition of code. Currently, there are quite a few classes that are identical to the old code. In these cases, we should aim to reuse the old classes. From my observations, the following classes are entirely identical:
```
	IDAQConfig2019        << >> IDAQConfig
	AbstractConfig2019    << >> AbstractConfig
	ESBCutConfig2019      << >> ESBCutConfig
	LBOCutConfig2019      << >> LBOCutConfig
	UBOCutConfig2019      << >> UBOCutConfig
	ULBCutConfig2019      << >> ULBCutConfig
	PairTriggerConfig2019 << >> PairTriggerConfig
```
We should remove the "2019" versions and just use the old ones, unless there is something different that I missed. Otherwise, we're maintaining two identical instances of the exact same class and code.

A number of other classes are very similar, but with some upgrades/changes. Ideally, we want to handle this by extending the old classes and only add new code where it is needed. This greatly improves maintainability. I will have to address these classes separately so that I can more specifically address what approaches we can take.

tongtongcao:
> HodoEcalCoincidence2019

tongtongcao:
> I'm still working on my full review, but here are a few quick comments.
> 
> **HodoEcalCoincidence2019**
> What information is this class supposed to store? Presently, it does not actually seem to define either mutator or selector methods for any values. I assume that it is intended to store whether or not a calorimeter/hodoscope coincidence is needed for a trigger. If that is the case, it would be better to write it as an `AbstractConfig<Boolean>` with a single parameter that is either `true` or `false`.
> 
> **TSConfig2019**
> I don't see any issues with the implementation itself. Everything looks good there! I think it would be useful, however, to include some public static final member variables to map what each index in the array corresponds to, so others who may work with the code can easily link the two. For example:
> 
> ```
> 		public static final int SINGLE_0_TOP = 0;
> 		public static final int SINGLE_1_TOP = 1;
> 		...
> ```
> 
> Then, when you get the array, you can just reference a given value with:
> 
> ```
> 		int[] prescales = tsConfig.getTSPrescales();
> 		int st0 = prescales[TSConfig2019.SINGLE_0_TOP];
> ```
> 
> This makes it so that the user doesn't need to directly reference the source code to know which array entry corresponds to what.
> 
> **PDECutConfig2019** / **FEETriggerConfig2019**
> This isn't strictly necessary to change, since it is really a bit pedantic of me, but Java standards generally make public static final variables in all caps. So, for instance, `parC0` --> `PAR_C0`. I'll still pass the pull request if you ignore this one, though. :)

Thanks for comments. Here are response.

For HodoEcalCoincidence2019, flags for Hodo and Ecal matching saved for single trigger. Refer to the codes in SinglesTriggerConfig2019.java:

     public SinglesTriggerConfig2019() {
        // Instantiate the base object.
        super(9);
        
        // Define the singles cuts.
        setValue(CUT_ENERGY_MIN, new LBOCutConfig2019());
        setValue(CUT_ENERGY_MAX, new UBOCutConfig2019());
        setValue(CUT_HIT_COUNT,  new LBOCutConfig2019());
        setValue(CUT_X_MIN,  new LBOCutConfig2019());
        setValue(CUT_PDE,  new PDECutConfig2019());
        setValue(L1_MATCHING,  new HodoEcalCoincidence2019());
        setValue(L2_MATCHING,  new HodoEcalCoincidence2019());
        setValue(L1L2_GEO_MATCHING,  new HodoEcalCoincidence2019());
        setValue(HODOECAL_GEO_MATCHING,  new HodoEcalCoincidence2019());
    }

You can see that we need a class inherited from AbstractConfig2019<Double> to store such flags. 

Refer to the codes in VTPConfig2019.java:
singlesTrigger[triggerNum].getL1MatchingConfig().setIsEnabled(parser.singlesL1MatchingEn[triggerNum]);       singlesTrigger[triggerNum].getL2MatchingConfig().setIsEnabled(parser.singlesL2MatchingEn[triggerNum]);          singlesTrigger[triggerNum].getL1L2GeoMatchingConfig().setIsEnabled(parser.singlesL1L2MatchingEn[triggerNum]);          singlesTrigger[triggerNum].getHodoEcalGeoMatchingConfig().setIsEnabled(parser.singlesL1L2EcalMatchingEn[triggerNum]);

You can see flags are saved.

For TSConfig2019, comments are nice. I have update codes.

mccaky (approved):
Code looks pretty good. It was decided that we want to keep the new trigger diagnostics separate from the old ones, so previous comments about re-using code are not relevant. Nothing obvious looks out of place, so merging is approved.

mholtrop (approved):

PR #654, opened 2020-02-04
change alignment histogram binning, ranges, cleanup

afilippi67:
Changes in appearance (binning, ranges, content) of the histogram set needed for alignment studies (valid for <=2016 and 2019 configurations). These histograms are only written in the GBLplots_ali.root file and not in the general output file (a cleanup is called to remove them from the heap).

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/gbl/GBLOutputDriver.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/gbl/GBLOutputDriver.java
@@ -31,16 +31,20 @@ import org.lcsim.event.base.BaseRelationalTable;
 import org.lcsim.geometry.Detector;
 import org.lcsim.util.Driver;
 import org.lcsim.util.aida.AIDA;
+import hep.aida.IManagedObject;
+import hep.aida.IBaseHistogram;
+
 
 /**
  * Make post-GBL plots needed for alignment.
  *
  * @author Miriam Diamond <mdiamond@slac.stanford.edu>
+ * @author Alessandra Filippi <filippi@to.infn.it>
  */
 public class GBLOutputDriver extends Driver {
 
-    public AIDA aida;
-    private String outputPlots = "GBLplots.root";
+    private AIDA aidaGBL; // era public 
+    private String outputPlots = "GBLplots_ali.root";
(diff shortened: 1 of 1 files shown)


normangraf (approved):
Looks good to me. Don't know how to test it.

tracking/src/main/java/org/hps/recon/tracking/gbl/GBLOutputDriver.java
```diff
@@ -31,16 +31,20 @@
...
  *
  * @author Miriam Diamond <mdiamond@slac.stanford.edu>
+ * @author Alessandra Filippi <filippi@to.infn.it>
  */
 public class GBLOutputDriver extends Driver {
 
-    public AIDA aida;
-    private String outputPlots = "GBLplots.root";
+    private AIDA aidaGBL; // era public 
+    private String outputPlots = "GBLplots_ali.root";
```
pbutti: Why change on the name of the outputPlots from default? It can be changed from the driver anyway. 
And I don't understand why public aida to private aidaGBL. Is there a motivation for changing type and name of the variable?

pbutti, line 47: Why change on the name of the outputPlots from default? It can be changed from the driver anyway.
And I don't understand why public aida to private aidaGBL. Is there a motivation for changing type and name of the variable?

pbutti (commented):
I'm ok with the rest of the changes

tracking/src/main/java/org/hps/recon/tracking/gbl/GBLOutputDriver.java
afilippi67: no, actually the new aida object is not needed - I replicated the aida object in the attempt to erase local histograms at the end to be sure not to spoil the rest of the stored histograms. I forgot to restore the previous situation upon committing (it doesn't make a difference, anyway), I will change it back and commit a new version.
The output filename is changed from the default because I prefer to have as default a name containing an underscore, for ease of subsequent analysis. Morevoer, not all of the plots are actually after GBL, so the older default naming could be misleading.

mholtrop (approved):

PR #656, opened 2020-02-13
Hodo developement

rafopar:
The change is tiny, so I haven't created a special issue for this.

I run the full integration test, no issues


rafopar:
Oops, I noticed that there were some old changes before, which were not merged with the master, 
however it was only in the hodoMonitoring application, which was tested and worked well during the summer's run.

So It is safe to merge it to the master.

baltzell (approved):

mholtrop (approved):

Issue #663, opened 2020-03-03
Add guard in AbstractSVTEvioReader to check Null pointer.


mholtrop:
This responds to #662 where Andrea first reported the problem.

For very rare events, hps-java crashes on an SVT hit that seems to correspond to a non existing channel. This then returns a Null pointer for the sensor in AbstractSvtEvioReader.java:300
`HpsSiSensor sensor = this.getSensor(data);`

Since this is rare, the proposed solution is to simply test for the Null pointer condition, and print a warning if this is encountered.

Issue #662, opened 2020-02-25
2019 reconstruction fails


andrea-celentano:
After a git pull and a clean recompile (commit b65e7daf442481bc4ed73b391449c28a2bb1d22a (HEAD -> master, origin/master, origin/iss204e, origin/HEAD, iss660)

I tried the command: 

java -cp ~/work/hps/hps-java/distribution/target/hps-distribution-4.5-SNAPSHOT-bin.jar org.hps.evio.EvioToLcio -r -x /org/hps/steering/recon/PhysicsRun2019_NoSVT.lcsim -d HPS-PhysicsRun2019-v1-4pt5 -DoutputFile=test hps_fee_010022.evio.00100-00108

The data file is available at www.ge.infn.it/~celentan/data

The error I get is:

Exception in thread "main" java.lang.NullPointerException
	at org.hps.evio.AbstractSvtEvioReader.makeHit(AbstractSvtEvioReader.java:303)
	at org.hps.evio.Phys2019SvtEvioReader.makeHit(Phys2019SvtEvioReader.java:124)
	at org.hps.evio.Phys2019SvtEvioReader.makeHits(Phys2019SvtEvioReader.java:174)
	at org.hps.evio.AbstractSvtEvioReader.makeHits(AbstractSvtEvioReader.java:191)
	at org.hps.evio.LCSimEngRunEventBuilder.makeLCSimEvent(LCSimEngRunEventBuilder.java:189)
	at org.hps.evio.LCSimPhys2019EventBuilder.makeLCSimEvent(LCSimPhys2019EventBuilder.java:56)
	at org.hps.evio.EvioToLcio.run(EvioToLcio.java:608)
	at org.hps.evio.EvioToLcio.main(EvioToLcio.java:92)

normangraf:
Please try running with the production jar found at

https://srs.slac.stanford.edu/nexus/repository/lcsim-maven2-snapshot/org/hps/hps-distribution/4.5-SNAPSHOT/hps-distribution-4.5-20200218.214859-16-bin.jar

Please let me know if this works.

pbutti:
Hi guys, 
Not sure if related but I see tests failing on the current master. Here is a log of the test and error

Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 20.628 sec <<< FAILURE! - in org.hps.test.it.PhysRun2019ReconTest
[...]

PR #664, opened 2020-03-04
Iss #663 and #662 - Guards around bad SVT hit codes.

mholtrop:
This sets a null pointer guard around the problem reported by Andrea (iss #662). The null propagates so needs to be tested for in several other location. 

Tested on hps_010022.evio.00050 and hps_010022.evio.00107, both files have several thousand events with this error.

### DIFF
--- a/evio/src/main/java/org/hps/evio/AbstractSvtEvioReader.java
+++ b/evio/src/main/java/org/hps/evio/AbstractSvtEvioReader.java
@@ -299,14 +299,18 @@ public abstract class AbstractSvtEvioReader extends EvioReader {
         // Get the sensor associated with this sample
         HpsSiSensor sensor = this.getSensor(data);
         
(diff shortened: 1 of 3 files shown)


pbutti:
Hi Maurik, 

thank you for this change.
adding the protection looks good to me. However, do we understand why 
HpsSiSensor sensor = this.getSensor(data);
returns null? 

Cheers,
PF

mholtrop:
Hello PF,

I hope that Omar will look into this. He is aware of the issue, but I don’t know when he will have time for this.

Best,
	Maurik

pbutti (approved):

pbutti:
Fine for me, then

rafopar:
I din't have much to say, except the integration test run just fine

rafopar (approved):
Integration test run just fine.

As I am not expert in this part. I am just approving it.

PR #670, opened 2020-03-11
Track Data stores the track momentum too from the first TrackState. iss669

### LINKED ISSUE #669: Add Track momentum to track data
Add the track momentum to track data. 
- Would be good to have the track momentum already stored in track data so we don't have to recompute that in hpstr passing the BField value as parameter
- In general useful if one wants to remake plots for tracks from LCIO. 
- Can be used in ReconParticleDriver to fill particle momentum correctly [TrackState.getMomentum() returns (0.,0.,0.) if running from reconstructed LCIOs files]. Will fix this in separate issue

pbutti:


mholtrop (approved):
Seems fine. All integration tests pass.

normangraf (approved):
Looks good. Would be nice to have a dedicated unit test for these quantities, but that shouldn't hold up this release.

PR #672, opened 2020-03-12
Iss671: Change http URLs to http

JeremyMcCormick:
Change http URLs to http in pom.xml files where applicable.

I only checked that the current master builds successfully using a fresh M2 repo. We will have to perform a release to fully test these changes.


mholtrop (approved):
Could use sed to do this change 😄

normangraf (approved):
What Maurik sed.
Builds and tests fine.
Approved.

PR #673, opened 2020-03-12
Iss666

mholtrop:
This cleans up the hodoscope code to output only one type of hit, instead of two. I dropped the generic hits, since all the information is available from the HodoCalHits.


baltzell (approved):

normangraf (approved):
Looks OK.
Release builds and tests successfully.
Would like to see more and better unit and integration tests, but that shouldn't hold up this release.

PR #676, opened 2020-03-17
Read 2019 RF signals

baltzell:
Note, need some restructuring on FADC evio readers (at least ECAL if not also HODO) to go kludgeless.

### DIFF
--- a/evio/src/main/java/org/hps/evio/EcalEvioReader.java
+++ b/evio/src/main/java/org/hps/evio/EcalEvioReader.java
@@ -256,6 +256,42 @@ public class EcalEvioReader extends EvioReader {
                 }
 
                 if (id == null) {
+                    //
+                    // FIXME:
+                    //
+                    // The original detector decision in makeHits above is attempting
+                    // to make it at the crate level.  Below it's generating
+                    // FADCGenericHit objects for any FADC channel that's not an ECAL
+                    // channel but shares an FADC crate with ECAL.
+                    //
+                    // The real crate number was also converted above to a "virtual"
+                    // crate number (1 or 2), but only if it's an ECAL FADC crate.
+                    //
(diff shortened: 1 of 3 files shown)


normangraf (approved):
Looks fine. Agree with FIXME comments in the code, but let's get this out now.

evio/src/main/java/org/hps/evio/EcalEvioReader.java
```diff
@@ -256,6 +256,42 @@ private static FADCGenericHit makeGenericRawHit(int mode, int crate, short slot,
...
+                    //
+                    // Starting in 2019, the RF FADC (in unshared slot 13) is now in
+                    // the same crate as ECAL.  RF software uses real crate number,
+                    // since that's what they previuosly received.
+                    // 
+                    // Note, HODO FADC hits are also converted here, but not used.
+                    //
+                    // Some options include, in order of goodness:
+                    // 1. restructure this reader (and maybe the HODO one too, or merge)
+                    //    (and put PMT/RF crate/slot/channel in conditions database)
```
mholtrop, line 282: 1++ We should probably re-do the way EVIO is read. The current structure for EVIO reading is in my view backwards, where the ECAL code goes "give me all the hits from this crate" and then tries to process the hits. Instead, I think we need to turn this around: get the hits, see who the hits belong to, send them to the appropriate code for processing.  Just my 2 cents on how to handle this.

mholtrop (approved):
Code looks good, but as Norman already pointed out, we need some detailed testing.

evio/src/main/java/org/hps/evio/EcalEvioReader.java
baltzell: Yes, we should, since 2015:)

Regarding the kludge and testing, note that slot-13 has never been used in HPS FADC crates before.

PR #685, opened 2020-03-30
Iss677 - Layer 0/1 changes and 2019 trigger drivers

### LINKED ISSUE #677: Validation 2019 Readout and Reconstruction Pipeline
The branch associated with this issue will be used to validate the updates needed for L0/1 of the SVT and the trigger sim.

omar-moreno:


normangraf (approved):
Builds and tests fine, so at least it doesn't break anything (that we're testing...). There are a LOT of changes in this request so I think we should continue performance testing after the merge.

pbutti (approved):

pbutti:
I cannot speak much about all the trigger part. For the recon part things look ok. Only doubt is the current settings for the "firstLayer" in the KalmanPatRecHPS that is now set to "0", probably breaking 2016 proper pattern reco at the moment.

normangraf:
Layering starts at 1.

pbutti:
It's a different numbering what I'm talking about.
In the KF pat reco, numbering starts at 0 and ends to 13.

PR #689, opened 2020-04-06
Update PhysicsRun2019MCRecon.lcsim

pbutti:
Reconstruction picks 2016 beamEnergy and the StandardCuts need to be overrided by steering file (otherwise 2019 projections won't be informative). 
- This PR adds a patch to the PhysicsRun2019MCRecon.lcsim to override wrongly defaulted mouse cuts.


omar-moreno (approved):

normangraf (approved):
Looks good.

PR #710, opened 2020-04-27
Disabled mouse cuts in both GBLRefitterDriver and ReconParticle Drive…

### LINKED ISSUE #709: remove mouse cuts.
And enable them from steering config file

pbutti:
…r. They can be restored by steering file


tongtongcao (approved):

mholtrop (approved):
Changes are fine. 
It would be great if we can do better with the documentation of the various constants in the steering files.

PR #722, opened 2020-07-03
Fix setting of 'notes' field in conditions CL tool

JeremyMcCormick:
Apache CLI does not appear to handle quoted arguments correctly, so the notes were not being added correctly. I changed this so that the extra args when using the command are used instead e.g.

``` 
add -c 4 -t svt_gains -r 1190000 -e 1200000 -u jeremym MC 2019
```

### DIFF
--- a/conditions/src/main/java/org/hps/conditions/cli/AbstractCommand.java
+++ b/conditions/src/main/java/org/hps/conditions/cli/AbstractCommand.java
@@ -106,7 +106,7 @@ abstract class AbstractCommand {
     /**
      * Print the usage of this sub-command.
      */
-    protected final void printUsage() {
+    protected void printUsage() {
         final HelpFormatter help = new HelpFormatter();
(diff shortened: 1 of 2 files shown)


baltzell (approved):
Are these "extra" args are always the last args?

JeremyMcCormick:
It would probably use all the extra rags you gave it but you should put
them at the end (similar to using a command like 'git stash save')

baltzell:
Ok, I was just wondering if, for example, incorrect intermediate arguments will now get interpreted as comments and end up doing weird/unexpected new stuff, rather than erring.  Anyway, not a big deal.

JeremyMcCormick:
Yes, probably incorrect args will be picked up as comments, but I don't
know a good way to avoid it.

JeremyMcCormick:
Behavior seems to occur in WSL and not at JLAB so I will avoid doing conditions stuff in WSL.

PR #746, opened 2020-10-31
updates and corrections relative to singles3 trigger

tongtongcao:
After deep discussion and double check with Ben, there are some updates and corrections relative to singles3 trigger in the readout system. 
1) Geometry mapping between hodoscope and ecal changes a little bit before and after run 10189 for 2019 experiment. The setup after run 10189 is regarded as standard and  updated in the system.
2) Fix scaling for hodoscope hits. For gains in the DAQ configuration used by the trigger system, they have been scaled by 1.25/2 for hodo channels at tiles with two holes. Instead, such gains have not been scaled in DB.
3) For hodoscope pattern bits: CL_12, CL_23, CL_34, and CL_45, it requires that there is at least one hits at BOTH left and right tiles.

Overall, the updates of items 1 and 2 have ignorable effects on readout, and the setup of item 3 causes a little bit tighten to produce single3 trigger. According to tests, # of triggered events with the new setup of the singles3 trigger is a few percents less than the old setup.

### DIFF
--- a/ecal-readout-sim/src/main/java/org/hps/readout/hodoscope/HodoscopePatternReadoutDriver.java
+++ b/ecal-readout-sim/src/main/java/org/hps/readout/hodoscope/HodoscopePatternReadoutDriver.java
@@ -68,11 +68,6 @@ public class HodoscopePatternReadoutDriver extends ReadoutDriver {
     private double hodoHitThreshold = 200.0;
 
     /**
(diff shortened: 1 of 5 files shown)


JeremyMcCormick (approved):

normangraf (approved):
Builds and tests cleanly. Please proceed.

PR #750, opened 2020-11-17
Iss742: Add condition for SVT sync status

### LINKED ISSUE #742: Add condition for SVT sync status
table: svt_sync_statuses
field: good

If good is true then the corrections do not need to be applied. 

False indicates corrections are needed.

----

List of good and bad runs here:

https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=295094627

JeremyMcCormick:
The SVT sync status condition is represented by the class `SvtSyncStatus` with the method `isGood()` returning `true` if the status is good and `false` if not.

Data is contained in the table `svt_sync_statuses` in the primary JLAB conditions database and was loaded from the information provided by @normangraf (see comments on the github issue).

There is a test that performs a spot check on a few of the records, which can be run as follows:

```
mvn test -Dtest=SvtSyncStatusesTest
```

This successfully checks that the first and last conditions records are correct (good and bad, respectively).


normangraf:
RawTrackerHitFitterDriver.java fails the style checker and does not compile. Please fix.

tracking/src/main/java/org/hps/recon/tracking/RawTrackerHitFitterDriver.java
```diff
@@ -183,9 +194,17 @@ public void process(EventHeader event) {
                     fit.setT0(fit.getT0() - timingConstants.getOffsetTime());
                 }
                 if (subtractTriggerTime) {
-                    double tt = (((event.getTimeStamp() - 4 * timingConstants.getOffsetPhase()) % 24) - 12);
-                    if (debug)
+                    double tt = (((event.getTimeStamp() - 4 * timingConstants.getOffsetPhase()) % 24) - 14);
```
mholtrop: This looks like it changes the tt offset by 2 from what was done previously. I would expect at least some comment why this new behavior is superior than what was done before.

cbravo135: This was discussed in multiple meetings. This shifts the average offset to be roughly zero, rather than the -2 it used to be. The intention was always to have a average time of roughly zero, and not -2.

cbravo135: I would like to see a comment as to why 12 was chosen in the first place, and why it wasn't immediately fixed when it was seen that the behavior of the code was not what was actually intended.

normangraf:
All of the 2015 & 2016 integration tests fail. This should not happen, as this issue was intended to resolve problems specific to the 2019 data.

tracking/src/main/java/org/hps/recon/tracking/RawTrackerHitFitterDriver.java
cbravo135: When I brought up it being at -2, it was argued it didn't matter because it changes the times the same amount and we only use relative hit times, not the absolute times, in our selection. I would say the same argument applies to this scenario, and zero is better because then defining a cut on the absolute time becomes a more simple cut.

cbravo135:
@normangraf this is an inaccurate representation of the intention of this issue.

cbravo135:
The intention was to improve the calibration of the hit time corrections.

JeremyMcCormick:
Can you paste the log from the test failures here?

cbravo135:
https://srs.slac.stanford.edu/hudson/view/HPS/job/hps-java-dev/62/testReport/

normangraf:
Failed tests:
  EngRun2015FeeReconTest.testIt:67->comparePlots:106 expected:<1> but was:<103>
  EngRun2015MollerReconTest.testIt:62->comparePlots:101 expected:<417> but was:<389>
  EngRun2015V0ReconTest.testIt:66->comparePlots:105 expected:<1949> but was:<1957>
  PhysRun2016FeeReconTest.testIt:62->comparePlots:89 expected:<125> but was:<128>
  PhysRun2016MollerReconTest.testIt:66->comparePlots:93 expected:<2221> but was:<2182>
  PhysRun2016V0ReconTest.testIt:66->comparePlots:93 expected:<4890> but was:<4743>

JeremyMcCormick:
This must be a consequence of changing the raw tracker hit fitting. Just
adding the sync condition would not do this.

normangraf:
That's correct. Just adding the SVT conditions should have no effect.

cbravo135:
If you look in jenkins, you can see that if I change the 14 back to 12, it passes everything.

cbravo135:
This is the same code, just with the 14 changed to a 12: https://srs.slac.stanford.edu/hudson/job/hps-java-dev/61/

normangraf (dismissed):
Please revert the offset to 12, after which we can approve this pull request and merge with master. If needed, we can open a new issue to deal with the question of whether the offset should be 12 or 14.

cbravo135:
We already did discuss changing this to 14 and the analysis group agreed the intention of this number is to put the average hit time at zero. The offset should be 14, and always should have been 14.

cbravo135:
Mathematically, it makes sense for it to be 14 since the trigger fw and code is in essence "rounding up" the time to the next trigger clock cycle, so it will be off by half of a 4 ns cycle (2 ns) plus the time to the center (from the later edge) of the svt trigger window which is 24 ns wide, so 2 ns plus 12 ns is 14 ns.

cbravo135:
12 is a bug, and I don't want to change my code to have a bug over not having one.

cbravo135:
Seems like it is the tests that should change, not the 14.

mholtrop:
I agree, the test should change. However, we cannot merge this until that is done. Can the test change be part of the pull request, or is that separate?

cbravo135:
The tests can change as part of this PR. I just don't know how to do it myself, so someone that understands how the test framework works would need to do it.

conditions/src/main/java/org/hps/conditions/svt/SvtSyncStatus.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Table(names = {"svt_sync_statuses"})
+public final class SvtSyncStatus extends BaseConditionsObject {
+
+    public static class SvtSyncStatusCollection extends BaseConditionsObjectCollection<SvtSyncStatus> {
+    }
+
+    @Field(names = {"good"})
+    public Boolean isGood() {
+        return this.getFieldValue(Boolean.class, "good");
```
cbravo135: This throws an exception: 
```java.lang.ClassCastException: Cannot cast java.lang.Integer to java.lang.Boolean```

cbravo135:
So, after reading the code for the Recon tests that are failing, we just need to update some of the reference files in: http://www.lcsim.org/test/hps-java/referencePlots/
Not sure who has perms to do this, but I have no idea how to upload anything there.
It does seem that there is no need to change any code to get the tests to pass, so we can safely merge this PR.

cbravo135:
The tests are now changed and legacy drivers were added for 2015/2016 recon.

cbravo135 (approved):

JeremyMcCormick:
My original PR added only a small bit of functionality to the conditions system. The relevant branch has now been used to add code that was far beyond the scope of what I originally intended, so I'm closing this. (Also, @cbravo135 please don't approve PRs where you are the primary author of most of the changes, even if they have been originally opened by someone else,)

cbravo135:
@JeremyMcCormick I was asked to review this PR by the software coordinator. Please don't close PRs without talking to the people involved first.

JeremyMcCormick:
@cbravo135 If you want this branch merged then open a PR using your account, not mine. DO NOT REOPEN THIS AGAIN.

JeremyMcCormick:
@cbravo135 You reviewing this PR is nonsense since you are the author of most of the changes.

cbravo135:
The fact of the matter is this PR is still small. Please tell that to @normangraf who directly asked me to provide a review.

mholtrop (approved):

PR #753, opened 2020-11-26
Pass1 dev fix (Testing only - DO NOT MERGE)

JeremyMcCormick:


steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_recon.lcsim
```diff
@@ -364,16 +364,16 @@
             <beamSigmaX>0.05</beamSigmaX>
             <beamPositionY>0.04</beamPositionY>
             <beamSigmaY>0.020</beamSigmaY>
-            <beamPositionZ>-10</beamPositionZ>
+            <beamPositionZ>-7.5</beamPositionZ>
```
JeremyMcCormick, line 367: I don't believe you need to set this. The beam Z positions should be in the conditions database now.

PR #755, opened 2020-11-26
Extract files from pass1-dev_fix which resolve iss730

JeremyMcCormick:
It was too difficult to actually cherry-pick commits out of the 250+ on the dev branch, so I checked out the relevant files which resolve issue 730 (performance enhancements to ADC signal fitting).

This should be okay in terms of conflict resolution, as the files should be identical, though the commit history from the dev branch may be lost (not a big deal in my opinion).

**If we decide to merge `pass1-dev_fix` all at once, then this PR can be closed without merging.**


JeremyMcCormick:
@pbutti requested to close

PR #756, opened 2020-11-28
Iss752: Add driver for overlaying signal onto beam events

### LINKED ISSUE #752: Add driver for overlaying signal onto beam events
Instead of using external merge tools, create a Driver that overlays signal events from a side event stream onto beam events. (Discussed with @tongtongcao)

JeremyMcCormick:
This new driver can be used externally to generate a file with combined beam and signal events, replacing usage of the _FilterMCBunches_ and LCIO merge tools. This has several advantages, including the elimination of several steps in the job chain as well as no longer writing a large "spaced" file with the signal events before they are merged. This should reduce disk usage of MC jobs, as well as improve the overall runtime performance.

Signal events can be optionally rejected as they are read using filters on total ECAL edep and number of Hodoscope hits (more filters can easily be added, as needed, by implementing the `EventFilter` interface and adding the appropriate setter methods).

I have provided an example steering file to run only the overlay and produce a combined file, as well as a sample steering file for running the overlay inline with the readout.

I verified that the new scheme produces triggers in the readout simulation by using examples in hps-mc to generate the beam and signal (tritrig) files and then running them through the sample readout configuration.


mholtrop:
Are there any studies that compare this with the previous method?
It seems to me that before we change the way we do MC, we need to make sure this does not change the results.

JeremyMcCormick:
This doesn't change the way we do MC. It is just a utility for eventually
replacing one component of the procedures but it isn't being used in
production right now.

cbravo135:
Changes would have to be made to hps-mc to actually use any of this.

JeremyMcCormick:
This is implemented as a standard lcsim driver. So you would just use this
in a steering file and not include the "filter bunches" component. I
included a sample steering file for reference. You would also want to pre
filter the tritrig events in case you want to throw some of them out based
on energy cuts, etc.

tongtongcao:
Jeremy and I ever had a deep discussion about this driver. 
The driver can replace filter tools in hps-java and merging tool in LCIO, and combine two steps (filtering and merging) together.
I ever did a basic test by 2016 MC. With the same setup for filtering, # of triggers from readout by the old processing and by updated processing with application of this driver is reasonably close.

Generally, I would say that this driver works well. It can save space and time comparing to the old MC processing.
For normalization of MC data, it requires to run out of signal events with enough beam events for merging. 
If we run out of signal events before running through all the requested beam events, an exception is thrown from the driver, but the output file would still be written out. The exception is not easy to be avoided, but it does not affect MC processing.
Additionally, the driver does not work if we need pure signal samples.

Certainly, with application of the driver, readout steering files and hps-mc job scripts need to be updated. 
Before these updates, we need to determine how to fix the issue for production of pure signal samples. Still using the old processing? Applying a file with all empty events as beam for merging, or other ways? Choice of different ways affects how we do updates for readout steering files and hps-mc job scripts.

tongtongcao:
The driver works. I prefer to approve this PR, and to leave corresponding updates for readout steering files later.

tongtongcao (approved):

cbravo135 (approved):
I agree with TongTong

PR #758, opened 2020-12-03
Fixing BaseCluster position bug

### LINKED ISSUE #757: BaseCluster objects modify the cluster position
The code in ReconParticleDriver.java uses the following instruction:

     cluster = new BaseCluster(cluster);

while this seems to be very "innocent", actually the BaseCluster object "cluster" obtained in this way as a weird behavior: if one calls cluster.getPosition(), the code is not returning the position of the original cluster passed to the creator method, but performs again the calculation of the position, according to: 

http://www.lcsim.org/sites/lcsim/lcsim-event-model/xref/org/lcsim/event/base/BaseCluster.html
and
 http://www.lcsim.org/sites/lcsim/lcsim-event-model/xref/org/lcsim/event/base/TensorClusterPropertyCalculator.html

This can be fixed by doing:

cluster = new BaseCluster(cluster);
baseClus.setNeedsPropertyCalculation(false);

andrea-celentano:
Fixing issue 757

### DIFF
--- a/analysis/src/main/java/org/hps/analysis/MC/ReconParticleDriverForTrackRefit.java
+++ b/analysis/src/main/java/org/hps/analysis/MC/ReconParticleDriverForTrackRefit.java
@@ -478,7 +478,9 @@ public abstract class ReconParticleDriverForTrackRefit extends Driver {
                 //before calculating nsigma.  Default is don't use corrections.  
                 Cluster originalCluster = cluster;
                 if(useCorrectedClusterPositionsForMatching){
-                    cluster = new BaseCluster(cluster);
+                    BaseCluster clusterBase = new BaseCluster(cluster);
+                    clusterBase.setNeedsPropertyCalculation(false);
+                    cluster = clusterBase;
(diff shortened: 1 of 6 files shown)


baltzell:
Why make the extra object like

```
clusterBase = new BaseCluster(cluster);
clusterBase.setNeedsPropertyCalculation(false);
cluster = clusterBase;
```

rather than 

```
cluster = new BaseCluster(cluster);
cluster.setNeedsPropertyCalculation(false);
```

?

baltzell:
And I guess this means BaseCluster kind of has a bug, in that its copy constructor doesn't fully replicate all the properties.

andrea-celentano:
Reason is that cluster is from the abstract type Cluster, hence when you do

cluster = new BaseCluster(cluster)

you don't have access to the method setNeedsPropertyCalculation. At least, this is what my Eclipse was warning me..

baltzell:
Ah, ok, presumably Cluster inherits from BaseCluster, and you can cast it rather than make an temporary object, but anyway, I can't get to lcsim source currently

normangraf (approved):
Builds and tests cleanly. Please proceed.

baltzell (approved):

PR #768, opened 2021-01-19
Update the readout system for application of DAQ configuration

tongtongcao:
- The 2019 DAQ configuration system is updated to be applied into the readout system. Basically, all necessary 2019 DAQ versions (totally, 0.6 MB for space) are added into resources of hps-java. Run number in command is mapped to a corresponding DAQ version.
- The readout system is updated to apply DAQ configuration by the readout system.
- Nothing needs to be changed for commands in old scripts or hps-mc job system, and nothing will be changed for outputs with implementation of them.
- The steering files for 2019 readout have been updated to apply 2019 DAQ configuration. 
- The updated readout system is proved to works well by comparison of results before and after updates.
- See more details from 
[Readout Updates for Application of the DAQ  Configuration.pdf](https://github.com/JeffersonLab/hps-java/files/5837639/Readout.Updates.for.Application.of.the.DAQ.Configuration.pdf)


normangraf (approved):
Builds and tests successfully. Please proceed.

cbravo135 (changes requested):
Adding all of the configuration files for the DAQ as resources in the jar file goes against previously set (2016 and 2015) precedent on how these configurations files are handled in Java. Seems like this is a bigger change than required since the 2015 and 2016 DAQ config files are kept in the mysql database. If this is how we would like to do this now, we should try to keep the code consistent for all of our runs and include the migration of the old config files from the db to the jar resources as is done here for 2019. This PR could also change the 2019 case to follow the precedent already set. Having a different solution for different datasets is certainly will make future changes and maintenance of the code more difficult.

mholtrop (approved):

cbravo135 (approved):

PR #769, opened 2021-01-28
Iss731: Remove events expected to have APV25 readout crosstalk

### LINKED ISSUE #731: 2019 Data: Filter out "Monster Events" during EVIO to LCIO translation.

cbravo135:
This branch was merged with branch iss742. Adding new database table class and adding code for emulation of apv25 readout buffers.


cbravo135:
Curently this fails a test like this:
```Results :

Failed tests: 
  FEETupleDriverTest.testIt:19->TupleDriverTest.testIt:61->TupleDriverTest.CompareTextFiles:80 new text file is missing branch fspTrkExtrpXSensorStereoTopL9/D
```
On Slack, it seemed multiple people were okay with just removing this test. If @normangraf or @mholtrop can recognize this, I will go ahead and just do that. Quoting Matt G "I’m ok with getting rid of that test".

cbravo135:
FYI, I locally tried removing this test and it compiles and passes all tests with that change. I went ahead and pushed a commit to this branch which removes this test in hopes of expediting this.

bloodyyugo (approved):
Builds fine and tests are ok...is there anything else I need to do to check this?

cbravo135:
@bloodyyugo you can test the code directly by using the new steering file PhysicsRun2019FullRecon_SvtFilterTest.lcsim and running that on any file in hps_010031. I still don't have the ability to put these new db entires into the server at JLab, but there is an sqlite file at SLAC: hps3/users/bravo/db/hps_conditions_test.db

mholtrop (approved):
I think Cam's presentations have been quite convincing that this is functioning quite well will very low loss of useful events if any. All the tests pass when merged with master.

normangraf:
Does this code presume that all the events are sequential? Has this been tested on files of skimmed events where this is not the case?

cbravo135:
The code has not been advertised to work on skims. It was actually explicitly stated it won't work on skims, because it would be impossible to do this on skims. This is why the main recon steering file does not use it, and the person running the code should be smart enough to know when and when not to use it.

PR #774, opened 2021-02-09
Iss437: Move test files to JLAB

JeremyMcCormick:
This update incorporates the following changes:

- Copied all test files referenced in unit and integration tests from SLAC to a web directory at JLab
- Got rid of dependence on the old test-data module and copied relevant files to the new www dir
- Cleaned up and rearranged test utility classes; added static utility methods for downloading data files and reference plots
- Removed a number of unused test drivers from the `integration-tests` module
- Removed all hard-coded references to lcsim.org
- Removed some useless print outs in the recon ITs
- Eclipse formatter ate all the extra spaces in modified Java files (cosmetic only)

All tests passed when I ran locally.

I verified that there are no files downloaded to the cache which have an lcsim.org URL any longer.

New base URL for the test files: 

https://hpsweb.jlab.org/test/hps-java/

All data files are under the `data` directory. The inconsistent SLAC directory structure was flattened for simplicity, so the files are all located in that directory with no sub-directories.

All reference AIDA plots are in `referencePlots` maintaining the same directory structure from SLAC.

Any HPS collaboration member with access to the JLAB `hps` account can copy test files to the new directory. This is an improvement over the old system since only users with a SLAC account could do this before. Files added by users need to have the correct read permissions or they will not be accessible by the web server.


normangraf (approved):
After clearing my cache I tested this at JLab using:

> mvn clean install -T1C

All tests complete successfully.

cbravo135 (approved):
This is a nice reorganization that improves maintainability and opens the software to be more collaborative.

PR #778, opened 2021-02-10
Update maven.yml

pbutti:
remove running tests

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -21,5 +21,5 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
-    - name: Build hps-java and run tests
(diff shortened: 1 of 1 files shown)


JeremyMcCormick (approved):

normangraf (approved):
Running the tests takes too long.
Making sure that the tests run successfully to completion should be a part of the QA done before requesting a PR.

JeremyMcCormick:
A lot of long builds could also eventually incur charges on the
JeffersonLab account, which I think we do not want.

The 'dev' branch can also be used for CI/CD instead of master. If you merge
to this branch, Jenkins will start a build with it similar to when PRs are
merged so you can check changes there should you not want to run the tests
locally. There is just some level of coordination required with everyone so
people don't step on each other's changes there.

PR #780, opened 2021-02-12
Pass1 dev fix

pbutti:
Resuming pull request.


JeremyMcCormick:
Trivial change: 

Please revert `tracking/pom.xml` back to 2 space indents to match all the other POM files in the project.

JeremyMcCormick:
I noticed there are changes to the geometry classes for generating the Java & LCDD geometries. This implies that all effected LCDD files should be regenerated and ideally checked for consistency with MC to make sure they match with the in-memory Java geometry. (Last step would be quite involved though requiring reconstruction or at least sisim to be run on all these detectors.)

I saw that some of the 2019 detector had changes but I wasn't sure if all of them had their LCDD files fully regenerated, which is something that should probably be done...

JeremyMcCormick:
Also, how to run with JNA should be documented someplace, probably here:

https://confluence.slac.stanford.edu/display/hpsg/Running+HPS+Java

omar-moreno (approved):
This is ready to be merged.

JeremyMcCormick (dismissed):
see my comments

pbutti:
I've tested running with the current classes over old lcdd files and seems like things work consistently. I haven't tested the generation of old lcdd as reading them seems to be backward compatible. 
One thing that should be fixed (and I'll do it next week) is ensuring the backward compatibility with the front Uchannel (which in the old case used 00 as MPID, and now uses 80). 
For the rest, I checked that the final sensors positions are the same with respect the usage of old compacts. Let me know if a dump of the positions is required.

Running with JNA is only for alignment purposes and will be documented, but I don't think should block the merge. 

Regarding the pom.xml, my editor is setup on 4 spaces. if you have one setup on 2 please go ahead and fix it, otherwise I'll take a look next week. 

I'm adding some simple validation plots made on 2019 FEE MC and 2019 V0 MC. 
FEE: https://www.dropbox.com/sh/5orkk8pwrvmnur0/AADOf1_xYcc7Gvvg_f1CXKwja?dl=0
V0:  https://www.dropbox.com/sh/go8xaeya7k4p4c5/AAC4bw1O9TJVXcjlMwrE-hvZa?dl=0

omar-moreno:
> Trivial change:
> 
> Please revert `tracking/pom.xml` back to 2 space indents to match all the other POM files in the project.

Can you fix this @JeremyMcCormick

JeremyMcCormick:
> > Trivial change:
> > Please revert `tracking/pom.xml` back to 2 space indents to match all the other POM files in the project.
> 
> Can you fix this @JeremyMcCormick

Fixed and pushed

JeremyMcCormick:
> Running with JNA is only for alignment purposes and will be documented, but I don't think should block the merge.

Agreed

JeremyMcCormick:
> I've tested running with the current classes over old lcdd files and seems like things work consistently.

The concern here would be that if the in-memory geometry is altered with the changes not reflected in the LCDD files, then you could get a situation where there are hits outside of the geometry or in the wrong volumes. This would cause errors in MC processing. 

I'm mainly bringing it up because it is something that has bitten us in the past, and probably we would want to at least spot check some of the detectors for the problem.

I don't know which LCDD files your changes affect - could be only 2019 detectors would need to be checked (?).

JeremyMcCormick:
I'm going to clear my review and leave it to another person to provide the second one.

I agree JNA doc can be added later, and I reformatted the POM file.

I just wanted to make sure that potential problems of mismatch between the Java and LCDD files were on the radar, but the amount of potential testing required to fully verify these compatibilities is probably prohibitive w.r.t. this PR.

JeremyMcCormick:
@omar-moreno @robertprestonjohnson @pbutti @normangraf 

If there are any, can the outstanding github issues be identified that are addressed by this PR so that they can be resolved when it is merged?

cbravo135 (approved):
:)

PR #791, opened 2021-02-22
Iss790: Refactor track-cluster matching algorithm implementation

alspellm:
Generalized the organization of Track-to-Cluster matching algorithms to allow for selection of different class based algorithms in ReconParticleDriver by simply setting the algorithm class name in the steering-file.
This update allows us to develop and use new track-to-cluster matching in reconstruction as easily as possible.


normangraf:
Let's have a presentation on this at the 02/24 software meeting and then proceed.

JeremyMcCormick:
This is looking good!

Can you revert the change to `analysis/pom.xml` as I'm going to be putting up a PR soon that excludes all the analysis module tests?

JeremyMcCormick:
Please don't use printing functions for output. Instead use a logger.

Driver already has a logger defined:

```
getLogger().info("hello");
```

Or if you need to create a new logger for your class outside of a Driver then define it statically:

```
private static final Logger LOG = Logger.getLogger(MyClassName.class.getName());
```

...where `MyClassName` is the name of your actual class.

And then in your code you can access it from anywhere, even static functions:

```
LOG.info("hello logger");
```

If you have any questions about this let me know.

JeremyMcCormick:
This is somewhat nit picky but can you cleanup all the extra spacing in between and within your functions?

You might as well make it look neater since you've done such a nice job of refactoring the track-cluster matching here...

JeremyMcCormick:
@alspellm Can you change your debug print outs to use a logger? (I DM'd you on Slack about this.)

This will allow people to turn on debugging for the class without having to recompile the code.

JeremyMcCormick (changes requested):
Requesting the following changes:

- [x] revert `analysis/pom.xml` to master -> `git checkout master -- analysis/pom.xml`
- [x] cleanup extra spacing
- [ ] change debug print outs to use a logger instead of a boolean flag
- [x] use TrackClusterMatcher for your interface name instead of TrackClusterMatcherInter (won't block on this either but I think it would be nice)

I checked that the ITs all pass with these changes on the `dev` branch.

JeremyMcCormick:
I would consider not having an interface called `TrackClusterMatcherInter` because it just strikes me as awkward. Can you just call it TrackClusterMatcher? Generally, you can just use a plain name for interfaces. (The classes then might be called Base, Abstract, Default or whatever.)

Since anyone who uses the interface typically won't see the underlying class, you should just go with the most natural name for it.

If you need to refactor/rename some classes to achieve this, I'd say go right ahead.

JeremyMcCormick (approved):
Alic made the changes I requested. Thanks!

cbravo135 (approved):
Great looking refactor. We should make an issue to discuss how to deal with the legacy drivers that need the legacy style of driver to deal with that in a future PR.

Issue #773, opened 2021-02-07
Convert print outs to log messages


JeremyMcCormick:
All these messages should be using a logger:

- [x] ```PF::TEMPORARY:: Move the kinMounts to a 2019 detector build method```

- [x] ```GBLRefitterDriver::process() -- Aborted refit of track -- null pointer for newTrackTraj returned from MakeGblTracks.refitTrackWithTraj .```

- [x] ```StripMaker::makeHits : Electrodes are of SiStriplets```

- [x] ```HPSTracker2014v1JavaBuilder::getSupportDetecDetector:```

- [x] ```[SvtEventFlagger] svt_readout_sync_phases not found java.lang.RuntimeException: No conditions were found with key: svt_readout_sync_phases```

- [x] ```[RawTrackerHitFitterDriver] svt_sync_statuses was not found: java.lang.RuntimeException: No conditions were found with key: svt_sync_statuses```

- [x] ```HpsGblFitter: Invalid GblTrajectory -> skip```

- [x] ```KalmanDriver: the B field is assumed uniform? false
KalmanTrackFit2 153: Failed to filter at site 8 layer 10.  Ignore remaining sites
KalmanTrackFit2 153: Failed to filter at site 9 layer 13.  Ignore remaining sites```

Didn't change this one - controlled by debug flag
- [ ] ```StripMaker ...```

- [x] ```ClusterEnergyCorrection2019: Loading resources data:```

PR #798, opened 2021-02-24
Iss776: Cleanup log messages, exclude some tests & cleanup IT print outs

JeremyMcCormick:
Resolves this issue:

https://github.com/JeffersonLab/hps-java/issues/773

See the issue page for complete details.

Excludes a number of tests that were not making any assertions. All tests in the analysis module were excluded. Several ITs that didn't test anything were excluded. (They can be added back later if needed as the source files were kept.)

Removes non-essential print outs and event printing from recon ITs. Timing was being printed out but this is already printed by Maven.


cbravo135 (approved):

normangraf (approved):
I would prefer that we added some teeth to the tests so they actually assert correct performance, but I'm OK with excluding these "no-op" tests for now and opening new issues when/if we decided to improve our testing suite.

JeremyMcCormick:
Thanks for the approval.

I want to revisit the recon ITs at some point and refactor common behavior
into shared classes and methods. We could also add performance requirements
as well if you think that's important to test.

normangraf:
I saw the new issue. Will respond to that.

PR #814, opened 2021-02-27
Iss805: Refactor and cleanup recon ITs

### LINKED ISSUE #805: Refactor recon ITs
There is a lot of duplication in the recon IT code. 

They could be improved by:

- Creating a base class that has constructor arguments for common parameters such as steering file, plot driver class, number of events, input file, etc.
- Using single methods for running recon and generating plots
- Using a single method for plot comparison which could be overridden or customized as needed
- Using generic methods for retrieving data and ref plots

JeremyMcCormick:
Resolves #805 

- Created common class for all recon ITs and refactored them to use it
- Added base class for the ref plot drivers with shared functionality and made the existing drivers extend it
- Moved ref plot drivers into the test classes as nested classes and deleted the standalone drivers
- Stripped out unused variables and settings from the ref plot drivers
- Added optional per event timing check
- Added optional print out of 1D histogram stats
- Added optional print out of memory usage after running reconstruction
- Made tolerance of histogram comparisons settable per test via the constructor
- Fixed bug where AIDA plots in folders were not being compared
- Added comparison of 2D histograms
- Deleted several old and obsolete ITs
- Tweaked a few warning messages in the tracking
- Decreased default tolerance of plot comparisons to 1E-6

Checked that all tests pass with the current tolerance.

There was a single histogram being checked with a different tolerance, and I did not feel that it was necessary to keep this. It can be added back in if absolutely necessary.


integration-tests/src/test/java/org/hps/test/it/ReconTest.java
```diff
@@ -32,7 +32,7 @@
     /**
      * The default tolerance level for histogram statistics comparison
      */
-    protected static final double DEFAULT_TOLERANCE = 5E-3;
+    protected static final double DEFAULT_TOLERANCE = 1E-6;
```
cbravo135, line 35: This is unnecessary, lets leave it at at 0.5%.

JeremyMcCormick: Was 0.5% tolerance arrived at by any rigorous determination?

The idea was that the next PR I put up will allow users to replace the reference plots as needed within their dev branches, so it wouldn't be necessary to have such a generous tolerance on the histogram comparisons, since they wouldn't need to be updated at JLAB any longer.  

Going forward, I'd much rather developers be aware when they change the output at all and then update the files, rather than having a fairly generous tolerance on the test statistics which could mask a series of changes to the recon output. So with this in mind, I'd prefer having the requirement be a nearly exact match between the ref and test output.

Maybe we can discuss further in the software meeting to see if there's a group consensus about it? (The tolerance can be set per-test now as well.)

cbravo135: Was 1E-6 arrived at by any rigorous determination? 0.5% is NOT "such a generous tolerance". Half a percent is rather conservative still. There are checks in these tests which will still catch if something changes at all. The code is currently 0.5% and you are changing it, how about we have a meeting where people agree to change this? Seems good where it already was. The other option would be just to have a tolerance of zero, which is effectively what it seems you are arguing for. Half of a percent is roughly an order of magnitude lower than 1/sqrt(1000) and I believe the stats for the samples that go into these tests are at least 1000 events. 0.5% is roughly what you will expect the stats error on the means to be given what the sample sizes roughly are. The point is 1E-6 is clearly way smaller than what you even expect the stats error will ever be in these tests.

bloodyyugo: The correct number here is 0...

JeremyMcCormick: "Was 1E-6 arrived at by any rigorous determination?"

It is a small number that will show you when there is any change in the the test output rather than a fairly large tolerance, at least for certain statistics, that could unintentionally mask changes. I don't remember the current value being discussed really at all, and it was slipped in there within a PR that was ostensibly unrelated without much discussion about it, so I don't think there's really any reason to consider it sacrosanct. The purpose of the tests is showing you when there are any changes to the physics reconstruction output, so the tolerance should be as low as possible. Depending on the plot, 0.5% could be a significant deviation in the mean or RMS, so it is too large of a tolerance for all of the statistics checks to fulfill their overall function.

"how about we have a meeting where people to agree to change this?"

Yes, let's do that...

normangraf: I concur with @bloodyyugo. There is no good way to automate this. Any changes should trigger human intervention with a "physicists" eye to what has changed and why. We should discuss this at the next software meeting.

JeremyMcCormick: > I concur with @bloodyyugo. There is no good way to automate this. Any changes should trigger human intervention with a "physicists" eye to what has changed and why. We should discuss this at the next software meeting.

We should definitely discuss this in a group setting.

Let me try all the tests with 0 as the tolerance just as a point of reference. 

I believe that will fail currently...

JeremyMcCormick: Currently, the tests fail at less than roughly 1E-6 tolerance, which was why I committed this number. I can test with a lower tolerance if I update the ref files (but I will wait for a more thorough discussion before doing anything).

cbravo135: Since computers are incapable of exactly calculating numbers, and rounding errors can change slightly depending on computer science issues and not physics issues, I disagree whole heartedly. IIRC, this number has never been zero for the relevant assertions. Simple differences in rounding errors can cause this, explicitly the numbers can be different if the references are generated with one system architecture/OS and the tests where the comparison to those references has a different arch/OS. Zero is not the "correct" number because of this and other related reasons. The "CS" eye is more adapted to spotting these issues than a "physicists" eye. Small deviations in the track chi^2 distribution are expected purely by running the tests a different system.

JeremyMcCormick:
I actually wanted to explore this a bit and see what results we get on
different systems like Jenkins vs local. A lower tolerance than 5E-3 is
certainly achievable across platforms (1E-6 works fine), but I don't want
to get this PR hung up on a single change of this number, which is not the
most important update nor the focus of the refactoring and improvements. If
we want to revert to the "current" value and figure out what this number
should be set to later on, I have no problem with this.

integration-tests/src/test/java/org/hps/test/it/ReconTest.java
cbravo135: A real review process is how a collaborative sw effort should go about understanding and dealing with serious changes.

cbravo135:
The current ref plots were generated on ifarm at JLab, FYI. I agree that we just leave the number where it was for now and discuss to decide what we want to call a serious "physics" change vs some CS issue that we don't want to get hung up on. My point before was that using roughly the expected statistical uncertainty seems like a reasonable way (IMO) to determine this threshold.

integration-tests/src/test/java/org/hps/test/it/ReconTest.java
```diff
@@ -0,0 +1,379 @@
...
+    }
+
+    /**
+     * Compare the number of entries, mean and RMS of two 2D histograms
+     * and assert they are equal within the configured tolerance
+     * @param refHist The reference AIDA tree
+     * @param testHist The test AIDA tree
+     */
+    protected void compareHistograms(IHistogram2D refHist, IHistogram2D testHist) {
+        assertEquals("Number of entries is different for: " + refHist.title(), refHist.entries(), testHist.entries());
```
cbravo135, line 302: Just want to point out that we do have assertions already that have ZERO tolerance which should be able to catch any serious change to the numbers given the ad hoc selection used for these tests.

normangraf (approved):
Please proceed.

bloodyyugo (approved):

PR #815, opened 2021-03-03
Update lcsim version

JeremyMcCormick:
Update lcsim to 4.3-SNAPSHOT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <dependency.locations.enabled>false</dependency.locations.enabled>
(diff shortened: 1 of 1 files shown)


cbravo135:
Just want to get a clarification on this. If one were to checkout some other version of lcsim locally, build, and install it, then it will still be whatever version was checked out there or will this override that somehow?

normangraf (approved):

JeremyMcCormick:
Maven will look for the lcsim version specified in the pom file when it
builds hps-java. It will first look in the local cache, and if there is a
matching version, it will use that jar. If it can't find the jar locally,
then it will download it from the remote repository.

So if you have a local lcsim 4.3-SNAPSHOT that you have installed, then
Maven will use it with whatever changes it might include.

Does that answer your question?

cbravo135:
Thanks, I think this does answer my question.

cbravo135 (approved):

PR #816, opened 2021-03-04
Change the default tolerance to not be senstive to floating point issues

cbravo135:


JeremyMcCormick (approved):
I am pretty well convinced that this fixes the issue with the intermittent test failure in Jenkins on master. I tried 3 builds in a row on the dev project with this tolerance and the tests passed.

JeremyMcCormick:
Here are the dev builds (73-76) with this change:

image

cbravo135:
This is currently failing on both my laptop and on SDF using 1E-4. Just pushed a commit to change it to 1E-3 which does not fail on my laptop or on sdf.
```
Failed tests: 
  EngRun2015V0ReconTest>ReconTest.testRecon:377->ReconTest.comparePlots:252->ReconTest.compareTrees:272->ReconTest.compareHistograms:292 RMS is different for: V0 Vertex z expected:<4.4841913337073655E-7> but was:<4.483562811546635E-7>
  ```

JeremyMcCormick:
Should we consider just getting rid of this plot in the ref set since it is
causing so many problems with floating point precision?

I don't see the same issues with the other plot statistics and setting the
tolerance for all the tests based only on this one plot having problems
doesn't seem like the best solution to me now...

cbravo135:
The whole scheme behind how the tests are implemented is the cultprit here. This problem might look like only one plot now, this won't necessarily always be true as we make more changes going forward. Just getting rid of the problem plot is just sweeping the real issue under the rug just as much as increasing the tolerance is.

bloodyyugo (approved):

PR #819, opened 2021-03-06
Fix an issue in the digitization readout driver

tongtongcao:
After updates with application of DAQ configuration for 2019 MC, pedestal from DAQ configuration is forced to be applied in the digitization readout driver. It causes that the readout system cant not work well for 2016 MC since pedestal for 2016 MC is obtained from database. With the update for the issue, the pedestal is optionally obtained from DAQ configuration or database. It depends on if we apply the DAQ configuration driver at the beginning of the readout steering file.  For the current readout system in hps-java, pedestal is from DAQ configuration for 2019 MC, while pedestal is from database for 2016 MC.


JeremyMcCormick (approved):

cbravo135 (approved):
I have run a simp sample locally to test this. I am getting triggers now, thanks for the quick turn around on this @tongtongcao !

PR #821, opened 2021-03-10
Disable checks of recon IT runtime

JeremyMcCormick:
Remove runtime checks of recon ITs.

I accidentally branched from dev but I think it should be fine. If we need a clean branch from master let me know...


cbravo135 (approved):
This passes tests on my laptop after I merged with master.

tongtongcao (approved):
Test passed in my machine as well.

PR #822, opened 2021-03-11
Hotfix for comparing histograms to remove issues with floating point

cbravo135:
errors and tolerances

### DIFF
--- a/integration-tests/src/test/java/org/hps/test/it/ReconTest.java
+++ b/integration-tests/src/test/java/org/hps/test/it/ReconTest.java
@@ -287,9 +287,10 @@ public abstract class ReconTest extends TestCase {
      * @param testHist The test AIDA tree
      */
     protected void compareHistograms(IHistogram1D refHist, IHistogram1D testHist) {
-        assertEquals("Number of entries is different for: " + refHist.title(), refHist.entries(), testHist.entries());
-        assertEquals("Mean is different for: " + refHist.title(), refHist.mean(), testHist.mean(), tolerance * abs(refHist.mean()));
-        assertEquals("RMS is different for: " + refHist.title(), refHist.rms(), testHist.rms(), tolerance * abs(refHist.rms()));
+        assertEquals("Number of bins is different for: " + refHist.title(), refHist.axis().bins(), testHist.axis().bins());
+        for(int ix = 0; ix < refHist.axis().bins(); ++ix){
+            assertEquals("Bin " + ix + " is different for: " + refHist.title(), refHist.binEntries(ix), testHist.binEntries(ix));
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


JeremyMcCormick (changes requested):
This is not the method of comparison that was discussed in the software meeting in which you were present. You argued for this "bin by bin" comparison, and it was not accepted. We are going to use histogram subtraction and produce comparison plots, and it was discussed that Norman and I would do it. This should not be merged.

cbravo135:
This is a completely unfair representation of the discussion had during the meeting. Your method is a bin-by-bin comparison as well. Everyone agreed checking the histograms were identical was the desired behavior of the assertions for this part of the tests. This should be merged so we have the desired behavior while you and Norman work on another implementation of the same behavior which will also include an easy to open file with comparison plots.

pbutti (approved):

bloodyyugo:
I don't see the issue with this...we can still make comparison plots, but this seems to do exactly what we want as far as the test goes.

JeremyMcCormick:
> I don't see the issue with this...we can still make comparison plots, but this seems to do exactly what we want as far as the test goes.

Going to wait to hear from @normangraf on this one. I'll remove my block once he's approved it.

normangraf:
This is one way to perform a bin-by-bin comparison, but is not the solution that was agreed upon at the 3/10/21 meeting. It also does not "fail fast" and removes useful information about how the histogram comparison is failing. I suggest we wait until we have a better solution that addresses all the issues discussed in the meeting.

cbravo135:
I am not sure what you mean it does not "fail fast" and how it removes any information? It tells you if a bin of the histogram is different, which is what was discussed in sw meetings. The statement "It also does not "fail fast" and removes useful information about how the histogram comparison is failing." is not true IMO. Matt G also agreed that "this seems to do exactly what we want as far as the test goes." In what way is that inaccurate?

cbravo135:
I have not presented this as a full stop solution. It seems to me it would be desirable to get what we wanted the test to do, done, in a quick way until we have time to go through and make improvements on the system. Does this PR not take a real step in the correct direction?

JeremyMcCormick:
The plan I proposed in the last software meeting for the recon tests is:

1) Use histogram subtraction as the comparison: test - ref

2) Save these histograms to an AIDA file (and eventually write a PDF with the output)

3) Make a second pass over the resultant histograms and fail the test if there are differences; we should just be able to look at the mean of the result, since it should be ~zero if there aren't any differences between the two

Assuming this plan is acceptable and then implemented, this PR would be overridden in the near future. So I don't really see the point in changing what we have now, since it works fine with the current level of tolerance to floating point calculations, and this new bin-by-bin comparison would soon be obsolete and replaced, anyways.

cbravo135:
I don't think we can really claim anything was agreed on in a meeting without clear slides of what is the only acceptable path to achieve the goal. As Jeremy has written this, I question what he means by 3, this to me seems to be a different outcome to having something mathematically identical to a bin-by-bin comparison. The mean of a histogram with all entries equal to zero has an undefined mean, mathematically.

JeremyMcCormick:
> Does this PR not take a real step in the correct direction?

It seems to me that without actually producing the comparison histograms for inspection by the developer, this method provides much less easily interpretable information than differences in the mean or RMS. Reporting a single difference in a bin height cannot be easily correlated to a change in the reconstruction code, whereas a reported shift in the mean or RMS is at least somewhat interpretable in a statistical sense. And if you actually changed the recon, then presumably every bin would be different, yet the assertion will only report the first one it sees. So I agree with Norman that this new way removes information on how the histograms actually differ. I really think we need to be creating plots here if we are going in this direction of comparing the histograms bin by bin, or the failures become basically impossible to interpret.

JeremyMcCormick:
> I don't think we can really claim anything was agreed on in a meeting without clear slides of what is the only acceptable path to achieve the goal. As Jeremy has written this, I question what he means by 3, this to me seems to be a different outcome to having something mathematically identical to a bin-by-bin comparison. The mean of a histogram with all entries equal to zero has an undefined mean, mathematically.

If you subtract the same two histograms from each other, all the bin heights should be zero, so we can assert on this state. That's what I meant by point 3.

cbravo135:
It is not true that plots are not produced for comparison. I disagree that the information provided when a test fails is impossible to interpret. This PR is mathematically identical to "If you subtract the same two histograms from each other, all the bin heights should be zero" as you just stated what you meant by 3. Just was pointing out that this is not a comparison of means, it will still be a bin-by-bin check of the histogram. How does this PR have any expectation to detect anything different from what you just said you meant by 3?

JeremyMcCormick:
I'm just arguing for status quo until we figure out what we're doing here and what the goal is, which may require additional discussion in a meeting (from the amount of discussion going on here I think this is probably the path forward). The plan that was discussed and tentatively agreed upon in the software meeting was not immediately throwing out the current set of test assertions and switching to a bin by bin comparison which throws cryptic errors about differences in bin heights. If you want some more details about what I'm suggesting we implement, including slides and a presentation in a meeting, I'm willing to provide it.

omar-moreno (approved):

Issue #796, opened 2021-02-24
Strip author tags from source files


JeremyMcCormick:
Remove all javadoc author tags from Java sources for the following reasons:

- Many of them include email addresses which cause errors in the Javadoc processing.
- Git keeps track of anyone who has modified a file in the log.
- The code is owned by the collaboration, not by individuals.
- Many of the files have been modified by two or more people which makes the tags inaccurate.
- Many of the source files already do not have the tags.

JeremyMcCormick:
Resolved by #824

PR #824, opened 2021-03-17
Iss796: Remove all author tags from source files

JeremyMcCormick:
See #796 for details.

Also, cleaned up class comments a bit in effected source files.

To see all users who have committed changes to a particular file use: `git log path/to/file`


pbutti (approved):

pbutti:
Hey Jeremy, out of curiosity: why did we remove the author-tags? 
I'll take care of not adding them to the next pushes! 

Cheers,
PF

JeremyMcCormick:
I listed the reasons in the linked issue.

JeremyMcCormick:
Main issue is that inclusion of email addresses in these tags causes the javadoc to generate an enormous number of errors. Also, everyone who has ever changed a file is listed in the git log, so that's a better way to find out who to blame. It is also extra boilerplate which just isn't needed and somewhat implies that HPS code is owned by individuals, which it isn't. Those are most of the reasons I can think of right now. :)

JeremyMcCormick:
The 'git log' command should tell you who created any file in the repo, and it should work even for old files that haven't been updated since the SVN days, since I mapped all the SVN users to github accounts when we imported.

JeremyMcCormick:
Watch about loading the file changes tab in your browser. It may cause it to freeze. :-p

analysis/src/main/java/org/hps/analysis/MC/FirstHitGBLRefitterDriver.java
```diff
@@ -254,4 +250,4 @@ static RelationalTable getHitToRotatedTable(EventHeader event, String RotatedHel
         }
         return hits;
     }
-}
\ No newline at end of file
+}
```
cbravo135, line 253: Looks like a closing curly brace was added without a counter opening one? I am probably missing something here.

analysis/src/main/java/org/hps/analysis/MC/KeepTruthRawTrackerHits.java
```diff
@@ -80,4 +78,4 @@ protected void process(EventHeader event) {
         }
         event.put(stripHitOutputCollectionName, truthStripHits, SiTrackerHitStrip1D.class, 0);
     }
-}
\ No newline at end of file
+}
```
cbravo135, line 81: Looks like this happens for a bunch of files.

analysis/src/main/java/org/hps/analysis/dataquality/MonsterEventAnalysis.java
```diff
@@ -183,4 +179,4 @@ public void process(EventHeader event){
             }
         }
     }
-}
\ No newline at end of file
+}
```
cbravo135, line 182: I really like all this clean up, but don't understand why these new lines with closing braces were added. If you help me understand this I am ready to approve.

cbravo135:
I really think this is probably me not understanding why github is marking it the way it is.

cbravo135:
Ok, talking with my room mate, pretty sure this means you removed the \n at the end of the files that had one. Looks good to me!

cbravo135 (approved):

JeremyMcCormick:
Okay, I think the newlines ARE supposed to be there. Do we care right now
that my vim ate them? I can fix it up later if needed....

PR #828, opened 2021-03-18
Archive legacy detectors.

### LINKED ISSUE #827: Archive legacy detectors
Most of the detectors in detectors-data/detectors are not actively used.  To improve the readability of the directory, legacy detectors are compressed into a tar.gz file. 

Another option is to just delete them. I don't think we need to go to that extreme yet ...

omar-moreno:
This resolves #827 . 

I'm sure we can add several of the 2016 detectors to the archive. Maybe others as well ...


JeremyMcCormick:
Can you restore this one for now?

HPS-EngRun2015-Nominal-v6-0-fieldmap_v3

It is used in the 2015 recon ITs.

pbutti (approved):

cbravo135:
I am okay with this if you were to remove the tests that are failing due to the "missing" detectors.

cbravo135:
Alternatively, you could put back a few of the detectors that were archived so the active tests will pass again.

JeremyMcCormick:
I would prefer the test detectors to be restored. I think deactivating a
bunch of recon tests is outside the scope of this PR.

normangraf:
We need to keep the detectors that are being used in tests. If we decide to delete those tests in the future we can delete the detectors as well.

JeremyMcCormick (changes requested):
restore detectors for recon tests as described in the comments please

JeremyMcCormick:
Closing this for now since it has been up over two months without being resolved.

PR #830, opened 2021-03-19
Iss811

### LINKED ISSUE #811: New TrackClusterMatcher compatible with KF and GBL
Creating a new TrackClusterMatcher algorithm, named TrackClusterMatcherMinDistance, based on closest distance matching

alspellm:
Added a new TrackClusterMatcherMinDistance.java along with some recon steering-files that use the new matcher.


omar-moreno:
@alspellm Can you post some basic plots showing residuals etc.?

alspellm:
Residual cuts are hard-coded at the moment:
dx < 20 mm
dy < 20 mm
dt < 8 ns


**Electrons**

image
image
image
image

**Electron Tracks that were NOT matched with a Cluster**
Tracks with P < 0.5 GeV aren't in the acceptance of the Ecal
image








**Positrons**

image
image
image
image


**Positron Tracks that were NOT matched with a Cluster**
Tracks with P < 0.5 aren't in the acceptance of the Ecal
image

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
```diff
@@ -0,0 +1,753 @@
...
+                }
+            }
+            Map.put(track, smallestdrCluster);
+        }
+        return Map;
+    }
+
+    private boolean isTrackInEcal(double trackx, double tracky){
+
+        //Define first order Ecal geometry --> assumed square here,
```
pbutti: Is it possible to read this numbers from geometry?

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
```diff
@@ -0,0 +1,753 @@
...
+    protected RelationalTable hitToStrips = null;
+    protected RelationalTable trackToData = new BaseRelationalTable(RelationalTable.Mode.ONE_TO_ONE, RelationalTable.Weighting.UNWEIGHTED);
+    protected String trackClusterCollectionName = "GBLTracks";
+    protected String rootFile = String.format("%s_TrackClusterMatching.root",this.trackCollectionName);
+    //Keep false. May be used in future 03.19.21
+    protected Boolean buildParamHistos = false;
+    /*
+     * Cuts used for track+cluster residual matching 
+     */
+    protected double tcut = 8.0;
```
pbutti: These should be configurable

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
```diff
@@ -0,0 +1,753 @@
...
+import org.lcsim.geometry.subdetector.HPSEcal3;
+import org.lcsim.event.ReconstructedParticle;
+
+
+/**
+ * Utility used to determine if a Tracks and Ecal Cluster are matched.
+ * Class extends AbstractTrackClusterMatcher, which implements
+ * TrackClusterMatcherInter
+ *
+ * @author <a href="mailto:alspellm@ucsc.edu">Alic Spellman</a>
```
pbutti: Remove

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
```diff
@@ -0,0 +1,753 @@
...
+            Map<Track, Map<Cluster, Double>> trackClusterResidualsMap = new HashMap<Track, Map<Cluster, Double>>(); 
+
+            for(Track track : tracks) {
+
+                 // Derive the charge of the particle from the track.
+                int charge = (int) Math.signum(track.getTrackStates().get(0).getOmega())*flipSign;
+
+                //Track time includes a hard-coded offset
+                double trackt = getTrackTime(track);
+                double tracktOffset = 3.5; //Track time distribution is centered on -4ns, added offset to center ~0
```
pbutti: Should be configurable

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
```diff
@@ -0,0 +1,753 @@
...
+
+                //Track positions depend on GBL or KF
+                List<Double> trackPos = this.getTrackPositionAtEcal(track);
+                double trackx = trackPos.get(0);
+                double tracky = trackPos.get(1);
+                double trackz = trackPos.get(2);
+
+                //dxoffset to be determined using truth information 03.19.21
```
pbutti: Same

pbutti (commented):
Why 2 steering files. If just for example, just do a minimal one.

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
alspellm: Not sure how to do this. This really isn't necessary to keep probably, I think I should remove. It was only really used as a sanity check

normangraf:
How does this matcher compare to the legacy matcher? It would be nice to see some plots for both MC and data.

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
alspellm: These are the offsets that were originally hard-coded that you asked me to just remove for now. Do you want these to be configurable from the steering-file for the time being?

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
alspellm: Can I add the dx and dy cuts in to the StandardCuts.java file?That seems like a reasonable way to pass these

pbutti: I find that class confusing. Avoid. Here is OK.

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
pbutti: Perhaps tracktOffset should be a member of the class more than a local variable, so then it's easy to configure from steering file.

recon/src/main/java/org/hps/recon/utils/TrackClusterMatcherMinDistance.java
pbutti: Perhaps isTrackInEcal should be a method of TrackUtils.

pbutti (approved):

cbravo135 (approved):
This code is mostly orthogonal to what exists, making the review more simple. It does change the NSigmaMatcher so I did locally run the tests and verify nothing was broken, though not expected from the changes made. Tests all pass, and since this is adding functionality and accepting this PR does not imply the analysis wg has accepted using this matching algorithm for use in any analysis, it is better to take this code now so we can collaboratively study this algorithm as a collaboration for use in a future analysis. Improvements and adjustments can still be made to this before it is used to officially reconstruct anything. Studies should be done before deciding to put this into the prod steering files, and since this PR does not, it appears safe.

pbutti:
Merged

PR #834, opened 2021-03-26
Hotfix for MC SVT hit times

cbravo135:
hotfix_mcSvtHitTime

Red is before this change, blue is after.

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019MCRecon.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019MCRecon.lcsim
@@ -89,6 +89,7 @@
             <subtractTriggerTime>true</subtractTriggerTime>           
             <!--per-strip timing correction from database...this should be on i f <useTimingConditions> is turned on in readout  -->  
             <correctChanT0>true</correctChanT0>          
+            <isMC>true</isMC>
             <debug>false</debug>
         </driver>
         <!-- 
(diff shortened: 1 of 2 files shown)


pbutti:
It's fine with me. Can you just maybe add a comment why this fix is necessary in MC in the code? 
Also do we expect this to be always to be applied also in future MCs? 

Finally, there are various tweaks that start to depend if an event is data or mc simulation and the options of the steering files are just starting to explode. We should think about a way to flag an event as a MC or Data so that the same algorithm can directly configure itself reading the eventHeader.

pbutti (approved):

cbravo135:
The relative phases of the trigger clock and svt clock can be different in different runs. Our DAQ did not consistently lock to the same relative phase during 2019 data taking, which is why the syncGood case is needed. I appears the MC "locks" to a different relative phase wrt either phase observed in 2019 data. The correction originally used assumes the trigger clock and svt clock are in phase, which only happens when syncGood = true in data.

cbravo135:
Since we use different steering files for data and MC, just setting isMC in the steering file is a reasonable solution, unless we decide at some point we would like to consolidate to using the same steering file for both data and MC.

cbravo135:
It is expected that this will always be the same relative phase in MC unless someone changes the phase in the readout code.

cbravo135:
The two peaks separated by 24 ns comes from them being out of phase. The correction puts the events, via the RF time, within 24 ns relative to the trigger clock wrt it being in phase with the svt clock. If they are not in phase, this will mean events are readout in different 24 ns svt clocks within the same 24 ns window of the trigger clock. Sorry if that is confusing, but this is a tricky thing to explain and I might not being doing a great job at it.

bloodyyugo (approved):

pbutti:
Using different steering files for Data and MC is what I'd like to avoid in the future.

PR #838, opened 2021-04-01
Fixed pass0 steering files

normangraf:
Fixed the situation where either the Ecal cluster corrections were being applied twice, or not at all.
Also modified target z position to reflect current guess of -7.5mm.

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_KFOnly.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_KFOnly.lcsim
@@ -124,7 +124,7 @@
             <beamSigmaX>0.05</beamSigmaX>
             <beamPositionY>0.04</beamPositionY>
             <beamSigmaY>0.020</beamSigmaY>
-            <beamPositionZ>-10</beamPositionZ>
+            <beamPositionZ>-7.5</beamPositionZ>
             <maxElectronP>7.0</maxElectronP>
             <maxVertexP>7.0</maxVertexP>
             <minVertexChisqProb>0.0</minVertexChisqProb>
@@ -132,7 +132,7 @@
             <maxMatchDt>40</maxMatchDt>
             <trackClusterTimeOffset>40</trackClusterTimeOffset>
             <useCorrectedClusterPositionsForMatching>false</useCorrectedClusterPositionsForMatching>
-            <applyClusterCorrections>false</applyClusterCorrections>
+            <applyClusterCorrections>true</applyClusterCorrections>
             <useTrackPositionForClusterCorrection>false</useTrackPositionForClusterCorrection>
             <debug>false</debug>
 	    <makeMollerCols>false</makeMollerCols>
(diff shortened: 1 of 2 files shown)


steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_KFOnly.lcsim
```diff
@@ -124,15 +124,15 @@
...
+            <beamPositionZ>-7.5</beamPositionZ>
             <maxElectronP>7.0</maxElectronP>
             <maxVertexP>7.0</maxVertexP>
             <minVertexChisqProb>0.0</minVertexChisqProb>
             <maxVertexClusterDt>40.0</maxVertexClusterDt>           
             <maxMatchDt>40</maxMatchDt>
             <trackClusterTimeOffset>40</trackClusterTimeOffset>
             <useCorrectedClusterPositionsForMatching>false</useCorrectedClusterPositionsForMatching>
-            <applyClusterCorrections>false</applyClusterCorrections>
+            <applyClusterCorrections>true</applyClusterCorrections>
```
pbutti, line 135: Hi Norman!
I think you set this to true here to pick the newest cluster corrections from Andrea, right?

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_recon.lcsim
```diff
@@ -337,7 +337,7 @@
             <maxMatchDt>40</maxMatchDt>
             <trackClusterTimeOffset>40</trackClusterTimeOffset>
             <useCorrectedClusterPositionsForMatching>false</useCorrectedClusterPositionsForMatching>
-            <applyClusterCorrections>true</applyClusterCorrections>
+            <applyClusterCorrections>false</applyClusterCorrections>
```
pbutti, line 340: But then I'm not sure I understand why it's set to false here for the particles formed by kalman tracks here but "true" for the particle formed by the older tracking (line 310)

bloodyyugo (approved):

cbravo135:
Can you show an E/p plot to demonstrate this does indeed fix the issue with the cluster corrections being applied twice?

normangraf:
TopBottomElectron_EoverP_20210405

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_recon.lcsim
normangraf: The correction needs to be applied only once. In the PhysicsRun2019_pass0_KFOnly.lcsim steering file we only construct ReconstructedParticles once, so the corrections need to be applied. In the PhysicsRun2019_pass0_recon.lcsim steering file, we create ReconstructedParticles twice, so the corrections should be applied only once, and since the KF ReconstructedParticles are being created after the GBL ReconstructedParticles, the corrections are turned off the second time through.

pbutti: Ok. I see

pbutti (approved):

PR #855, opened 2021-04-29
Add names of some missing developers to top level pom

cbravo135:


normangraf (approved):
Corrects an oversight.

mholtrop (approved):

PR #856, opened 2021-05-03
Update PhysicsRun2019_pass0_recon.lcsim

normangraf:
Increase maximum electron momentum from 7 to 10GeV to be included in FinalStateParticle collections. Until we fix the momentum reconstruction in the SVT top, cutting at 7GeV is too restrictive. Relax for now so we keep the higher momentum tracks in order to better understand what is going on.


tongtongcao (approved):
The setup makes sense. I approve it.

pbutti (approved):

PR #865, opened 2021-05-21
Iss864

normangraf:
I removed superfluous code in ReconParticleDriver that accessed the rotated helical track hits for no reason.
One can now run the Kalman Filter track finding and fitting without having to create HelicalTrackHits.
The [KF only steering file](https://github.com/JeffersonLab/hps-java/compare/master...iss864#diff-486de823c08371045aebe80b2e41e45995f8031b417d0fbfa0b1f97bdf615272) has been modified to remove that Driver.
It has also been updated to use the distance-based track cluster matcher.
It also now keeps monster events instead of zeroing out the 1D strip hit collection.

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_KFOnly.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_KFOnly.lcsim
@@ -111,7 +111,9 @@
           
         <driver name="ReconParticleDriver_Kalman" type="org.hps.recon.particle.HpsReconParticleDriver" > 
             <ecalClusterCollectionName>EcalClustersCorr</ecalClusterCollectionName>
-            <trackCollectionNames>KalmanFullTracks</trackCollectionNames>          
+            <trackCollectionNames>KalmanFullTracks</trackCollectionNames>  
+            <matcherTrackCollectionName>KalmanFullTracks</matcherTrackCollectionName>
+            <trackClusterMatcherAlgo>TrackClusterMatcherMinDistance</trackClusterMatcherAlgo>        
             <unconstrainedV0CandidatesColName>UnconstrainedV0Candidates_KF</unconstrainedV0CandidatesColName>
             <unconstrainedV0VerticesColName>UnconstrainedV0Vertices_KF</unconstrainedV0VerticesColName>
             <beamConV0CandidatesColName>BeamspotConstrainedV0Candidates_KF</beamConV0CandidatesColName>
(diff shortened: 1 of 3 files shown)


cbravo135:
Can you show some checks that this did not change the behavior of PhysicsRun2019_pass0_recon_evio.lcsim?

normangraf:
The two collections that were being accessed were never being used. Commenting out those calls should not affect the results of any other Drivers.

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_KFOnly.lcsim
```diff
@@ -87,7 +89,7 @@
        
         <driver name="TrackerHitDriver" type="org.hps.recon.tracking.DataTrackerHitDriver">
             <neighborDeltaT>8.0</neighborDeltaT>
-            <saveMonsterEvents>false</saveMonsterEvents>
+            <saveMonsterEvents>true</saveMonsterEvents>
```
cbravo135, line 92: Why is this changed from false to true?

cbravo135:
I think at a minimum you should run on one file and show that these collections are indeed removed as expected.

cbravo135:
Also, if we plan on moving forward with this steering file now, you should add the EventFlagFilter driver and turn on the "svt_readout_overlap_good" flag.

normangraf:
Seems to be a "Fitting with: Simplex" debug printout left in the code.


Ooops! Added this comment to the wrong PR.

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_KFOnly.lcsim
normangraf: I don't believe the current way of handling these events is correct, and simply skipping these events may be giving us a false impression of how well we're doing. I'd like to see what effect this has in combination with enabling the event flag filter.

normangraf:
The collections aren't (yet) being removed. I was simply removing an extraneous call to access those collections, which requires them to be in the event. Once we drop the SeedTracker  and GBL track finding and fitting these collections will no longer be in the event. No behavior has changed.

bloodyyugo (approved):

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass0_KFOnly.lcsim
cbravo135: The event flag filter is not currently activated in this steering file.

cbravo135: If this is your intention, please add it to this steering file and I will approve this PR.

pbutti (approved):

Issue #733, opened 2020-10-07
Implement remote AIDA support in online recon


JeremyMcCormick:
Use the freehep-jaida-remote package to implement remote filling of shared histograms by online reconstructions stations.

JeremyMcCormick:
Resolved by #868

PR #868, opened 2021-05-27
Online recon dev

JeremyMcCormick:
This is the currently working version of the online recon and monitoring tools, as described in several software meeting presentations, resolving #733.

These changes implement a server which is able to launch instances of the reconstruction which fill remote AIDA trees. These are aggregated and can be viewed in a web browser or JAS3.

A working test configuration can be found here:

https://github.com/JeremyMcCormick/hps-projects/tree/master/projects/online-recon-test

Documentation is on Confluence:

https://confluence.slac.stanford.edu/display/hpsg/Online+Reconstruction+Tools


normangraf (approved):
I've not explicitly run this, but the existing tests run to completion and it adds needed functionality.
Please proceed.

recon/src/main/java/org/hps/recon/filtering/EventFlagFilter.java
```diff
@@ -10,8 +11,9 @@
 public class EventFlagFilter extends EventReconFilter {
 
     private static Logger LOGGER = Logger.getLogger(EventFlagFilter.class.getPackage().getName());
-    
-    String[] flagNames = {"svt_bias_good", "svt_position_good", "svt_burstmode_noise_good", "svt_event_header_good", "svt_latency_good", "svt_readout_overlap_good"};
+    private static Level LEVEL = Level.WARNING;
+
+    String[] flagNames = {"svt_bias_good", "svt_position_good", "svt_burstmode_noise_good", "svt_event_header_good", "svt_latency_good"};
```
cbravo135: I think this will break the apv25 readout crosstalk filter.

cbravo135: It doesn't look like it does as this gets overwritten by the steering file. Is there a reason for removing "svt_readout_overlap_good" from this here though? I would prefer we just keep it so a list of all possible flags is kept here.

JeremyMcCormick:
Unintentional change - I will revert it.

recon/src/main/java/org/hps/recon/filtering/EventFlagFilter.java
JeremyMcCormick: This should be fixed now. I reverted the flag names array.

cbravo135 (approved):
Mostly additions to the code only slightly touching code that already existed. I also ran reco on a 2019 evio file and didn't see any obvious issues.

PR #869, opened 2021-05-30
Iss826: Move several utility modules into separate project

JeremyMcCormick:
I have moved the following modules into a separate project:

crawler
run-database
datacat

They are rarely used and do not need to be included in the primary repository.

The new project is at:

https://github.com/JeffersonLab/hps-utilities

The new project does not have automatically generated scripts or a bin jar, but this can be added easily, should someone want to use these tools.


normangraf (approved):
This is a useful cleanup, preserving  functionality while reducing clutter.
Please proceed.

bloodyyugo (approved):

PR #874, opened 2021-06-03
Iss872

### LINKED ISSUE #872: Enable second pulse fit to APV25 waveform to be dropped.
Currently we allow a second pulse to be fit to the SVT APV25 waveforms. Both fits are currently kept. This code would allow the fit farther from t0=0 to be dropped.

normangraf:
In cases where the APV25 waveform is fit to two pulses, this code only saves the fit whose time is closer to t=0.
Additionally, all fits with times less than a settable time _minT0_ or greater than a settable time _maxT0_ can be dropped by setting _dropOutOfTimeFits_ in the steering file.

This pull request produces some differences to three of the integration test plots, so will require the reference plots to be updated. 

Failed tests:
  EngRun2015V0ReconTest>ReconTest.testRecon:380->ReconTest.comparePlots:252->ReconTest.compareTrees:272->ReconTest.compareHistograms:293 Bin 99 is different for: V0 Vertex x expected:<63> but was:<64>
  PhysRun2016MollerReconTest>ReconTest.testRecon:380->ReconTest.comparePlots:252->ReconTest.compareTrees:272->ReconTest.compareHistograms:293 Bin 61 is different for: Moller Bottom Track Momentum expected:<119> but was:<118>
  PhysRun2016V0ReconTest>ReconTest.testRecon:380->ReconTest.comparePlots:252->ReconTest.compareTrees:272->ReconTest.compareHistograms:293 Bin 66 is different for: V0 Track Chisq Prob expected:<89> but was:<90>

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/RawTrackerHitFitterDriver.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/RawTrackerHitFitterDriver.java
@@ -1,6 +1,8 @@
 package org.hps.recon.tracking;
 
+import static java.lang.Math.abs;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import org.hps.conditions.database.DatabaseConditionsManager;
 import org.hps.conditions.svt.SvtSyncStatus.SvtSyncStatusCollection;
@@ -47,6 +49,10 @@ public class RawTrackerHitFitterDriver extends Driver {
     private double trigTimeOffset = 14.0;
 
     private double tsCorrectionScale = 240;
+
+    private boolean _dropOutOfTimeFits = true;
+    private double _minT0 = -50;
+    private double _maxT0 = 50.;
 
     /**
      * Report time relative to the nearest expected truth event time.
@@ -106,25 +112,27 @@ public class RawTrackerHitFitterDriver extends Driver {
     }
 
     public void setFitAlgorithm(String fitAlgorithm) {
-        if (fitAlgorithm.equals("Analytic"))
+        if (fitAlgorithm.equals("Analytic")) {
             fitter = new ShaperAnalyticFitAlgorithm();
-        else if (fitAlgorithm.equals("Linear"))
+        } else if (fitAlgorithm.equals("Linear")) {
             fitter = new ShaperLinearFitAlgorithm(1);
-        else if (fitAlgorithm.equals("PileupAlways"))
+        } else if (fitAlgorithm.equals("PileupAlways")) {
             fitter = new ShaperPileupFitAlgorithm(1.0);
-        else if (fitAlgorithm.equals("Pileup"))
+        } else if (fitAlgorithm.equals("Pileup")) {
             fitter = new ShaperPileupFitAlgorithm();
-        else
+        } else {
             throw new RuntimeException("Unrecognized fitAlgorithm: " + fitAlgorithm);
+        }
     }
 
     public void setPulseShape(String pulseShape) {
-        if (pulseShape.equals("CR-RC"))
+        if (pulseShape.equals("CR-RC")) {
             shape = new PulseShape.CRRC();
-        else if (pulseShape.equals("FourPole"))
+        } else if (pulseShape.equals("FourPole")) {
             shape = new PulseShape.FourPole();
-        else
+        } else {
             throw new RuntimeException("Unrecognized pulseShape: " + pulseShape);
+        }
     }
 
     public void setFitCollectionName(String fitCollectionName) {
@@ -143,12 +151,25 @@ public class RawTrackerHitFitterDriver extends Driver {
         this.rawHitCollectionName = rawHitCollectionName;
     }
 
+    public void setDropOutOfTimeFits(boolean b) {
+        _dropOutOfTimeFits = b;
+    }
+
+    public void setMinT0(double d) {
+        _minT0 = d;
+    }
+
+    public void setMaxT0(double d) {
+        _maxT0 = d;
+    }
+
     @Override
     public void startOfData() {
         fitter.setDebug(debug);
         fitter.setFitTimeMinimizer(fitTimeMinimizer);
-        if (rawHitCollectionName == null)
+        if (rawHitCollectionName == null) {
             throw new RuntimeException("The parameter rawHitCollectionName1 was not set!");
+        }
     }
 
     protected void detectorChanged(Detector detector) {
@@ -156,8 +177,7 @@ public class RawTrackerHitFitterDriver extends Driver {
         try {
             syncStatusColl = DatabaseConditionsManager.getInstance().getCachedConditions(SvtSyncStatusCollection.class, "svt_sync_statuses").getCachedData();
             syncGood = syncStatusColl.get(0).isGood();
-        }
-        catch (Exception e) {
+        } catch (Exception e) {
             syncGood = true;
             getLogger().config("svt_sync_statuses was not found.");
         }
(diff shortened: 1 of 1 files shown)


cbravo135 (changes requested):
As written this code change will always remove the pulse further from zero when two exist, and the old behavior is not possible. This has not been agreed on at any meeting, so the default should be the old behavior, which currently is not even possible. This explains why the tests are failing. Please make the code default to the old behavior of keeping both pulses when two pulses are fit to a waveform. It was actually discussed that more studies need to be done before considering the behavior this code explicitly forces.

bloodyyugo:
I'm a little hesitant to put a stopgap in now when we should just spend the effort to do this right.  By "right" I mean either fix the relations so that we can get both fits from the raw hit (and then decide in tracking) or we look at the other fit parameters (amplitude, uncertainties?) to help figure out which hit to keep (although, probably these will prefer the in-time hit).  I'm also afraid that these two-pulse hits aren't very good to begin with, but I'd like to keep that info (that they are two-pulse instead of one-pulse) somewhere where we can get it in analysis.  I don't think we are doing that now.

cbravo135:
As the code is currently, we do keep both pulse fits. No fix is needed to access them during tracking.

cbravo135:
A more flexible solution would be to add a driver to filter them before tracking.

normangraf:
This change was discussed at length in at least two presentations ([part 1](https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=311544084&preview=/311544084/311544674/20210525_SVTHitAnalysis_ReconMeeting_Graf.pdf)[part 2](https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=311544084&preview=/311544084/311544682/20210525_SVTHitAnalysisII_ReconMeeting_Graf.pdf) ) These changes were specifically proposed and no objections were raised at the time.

normangraf:
> I'm a little hesitant to put a stopgap in now when we should just spend the effort to do this right. By "right" I mean either fix the relations so that we can get both fits from the raw hit (and then decide in tracking) or we look at the other fit parameters (amplitude, uncertainties?) to help figure out which hit to keep (although, probably these will prefer the in-time hit). I'm also afraid that these two-pulse hits aren't very good to begin with, but I'd like to keep that info (that they are two-pulse instead of one-pulse) somewhere where we can get it in analysis. I don't think we are doing that now.

I drop poor hits, keep all the good hits, don't lose any tracks, speed up the reconstruction and make the output file size smaller. Calling this a "stopgap" trivializes the amount of work I have put into analyzing this issue and ensuring that the changes to the code do not result in any appreciable changes in our "good" events. No tracks were lost in either the 2016 Møller or V0 integration tests and at most a few tracks changed slightly (and perhaps even for the better). For your convenience the comparison plots for the 2016 V0 integration test can be found in the attached file.  
[plots.pdf](https://github.com/JeffersonLab/hps-java/files/6601639/plots.pdf)

cbravo135:
Objections to this were raised during both presentations cited. During the referenced discussions we agreed that we would like to further study the errors we gain to the hit times before making any decision about how to filter the events. We also did not discuss implementation details, and this code forces the proposed behavior while not allowing the current behavior. A consensus has not been made about this proposal and follow up studies were requested to further understand the ramifications of this change.

cbravo135:
It has not been demonstrated that this will "keep all good hits", and plots in the presentations cited suggest (but do not establish) that some good hits are dropped. A definition of a "good hit" has not been established in an attempt to quantify this.

mholtrop (approved):
Difference in the integration tests are completely minimal. We have approved changes with far larger differences without any fuss. This project was well discussed and presented at the software meeting, and this does not seem a stop-gap fix. 
There seems to me to be no reason to add complexity so that someone can restore the old faulty behavior. If you want that, you can just get the old version of the driver.

cbravo135:
The integration tests use highly skimmed data for clean events from 2016 and 2015. The tracking used is seed tracker. The discussion during the meeting indicated that we were first interested in studying the cases when two pulses are used and possibly one one really should be used. There is no reason this requirement must be in this driver, and for future studies it would be easier for this filtering to happen in a driver after this so the pulse fitting can still be studied fully after running. Tracks should be studied using the planned 2019 reconstruction configuration using a run that is not during the pulsed beam saga. The second set of slides is exclusively looking at a run that is not representative of running conditions for over 80% of our 2019 data.

normangraf:
This is a pull **request**.  I've demonstrated that we don't lose any "good" events, as previously defined, and the only changes over thousands of events was a slight change in the track chi-squared of a few tracks. Please feel free to proceed with your own studies to define what a "good hit" should be and to process representative runs in order to study the cases when two pulses are used and possibly only one really should be used.

cbravo135:
You have not attempted demonstrate we do not lose any "good" events in any way it could be defined in 2019, since you have not checked 2019 tracking using this code. You have not processed representative runs to study these cases where you claim to be able to throw away one pulse when two are used.

normangraf:
All of the plots in both of my presentations are from 2019 data. I did not "analyze" any 2015 or 2016 data until I ran the integration tests.

cbravo135:
You used run 10515 which is a run in 2019 with a known issue that is only present in about 12% of the 2019 data. It is not representative of the conditions of the 2019 run.

normangraf:
Please see my presentation at the June 8, 2021 analysis meeting:
https://confluence.slac.stanford.edu/display/hpsg/June+8%2C+2021+Meeting?preview=/311553517/311553773/20210608_SVTHitAnalysisIII_AnalysisMeeting_Graf.pdf
I've now shown that there is no loss of tracks when implementing this change in the 2015 & 2016 integration tests and in a sensitive three-prong trident analysis of the 2019 data spanning all runs. In addition, of the few tracks which are changed, the changes are minimal.

cbravo135:
This presentation does not describe the selection used for the three prong-tridents. I notice the rate of these are rather low. If we are going to force the behavior to be to drop one of the fits, there should be an independent check on a less cherry picked selection of events. "No loss of tracks" is too strong of a claim for what is presented in those slides.

normangraf:
Please proceed with your independent check. We can discuss it at the next analysis or reconstruction meeting.

cbravo135:
Please provide the provenance and details of the selection for the analysis shown in https://confluence.slac.stanford.edu/display/hpsg/June+8%2C+2021+Meeting?preview=/311553517/311553773/20210608_SVTHitAnalysisIII_AnalysisMeeting_Graf.pdf

cbravo135:
After digging a bit more, it turns out that dropping the fit further from t0 = 0 is already done in the clustering algorithm: 
https://github.com/JeffersonLab/hps-java/blob/master/tracking/src/main/java/org/hps/recon/tracking/NearestNeighborRMSClusterer.java#L137

This is why there is no change in tracking, because this cut here would not be expected to produce any different set of clusters. This only makes it more cumbersome to study the pile up fitting.

normangraf:
> After digging a bit more, it turns out that dropping the fit further from t0 = 0 is already done in the clustering algorithm:

Yes, indeed! And if you look closely you will notice that the out-of-time strip clusters are dropped as well, which is why I also added the ability to drop out of time fits. That is the whole point of this issue: to drop fits that will in any case never be used. By "failing early" it reduces both the processing time and the output file size without affecting the results in any appreciable way.

Please remove your block so we can proceed to the next stage of data reconstruction.

cbravo135:
This is not needed for the next stage of data reconstruction. It has not been demonstrated this gains us an appreciable amount of time or disk space. This will lose information about the fits performed so we can analyze them and work to improve the decisions about when to use one fit or two fits. When both fits are within 50 ns, we should consider adding the amplitudes during clustering if unable to remove all the fits when the second pulse is ahead in time.

cbravo135:
Currently the out of time cluster window is 144 ns wide, so we aren't dropping that many fits that way: https://confluence.slac.stanford.edu/display/hpsg/June+8%2C+2021+Meeting?preview=/311553517/311556300/anaMeet210608.pdf

cbravo135:
Please make it configurable so our collaborators can study the hit fit time errors we recently started using migrad to get.

cbravo135:
This PR as is loses information that could be analyzed by the SVT group, and has not been shown to gain anything appreciable.

pbutti:
Hi All, 
I don't fully understand why this is so controversial, but perhaps would be agreeable to use a flag that can be switched from steering files to activate Norman's selection of the fitted T0? 
I don't have anything against this PR, but if we want to keep things open for further study, and not everyone feels comfortable with this change, why not adding a boolean flag for activation at this stage?

cbravo135:
It is not true that this PR only "drop fits that will in any case never be used." as claimed by @normangraf in his last comment. Hit fits with a time of -70 ns would be dropped whereas they are currently used by code that runs after this in all current steering files with reconstruction of SVT clusters. @normangraf claimed "That is the whole point of this issue: to drop fits that will in any case never be used." and this PR is outside the scope of that claim.

bloodyyugo:
I agree with PF...let's add a flag so that this can be enabled/disabled via steering file and then I'm fine signing off.

normangraf:
That's a lot of extra work.
And I'm sure Cameron will argue to turn the switch off.

bloodyyugo:
I just looked at the code and you already have the setters there.  It look like you just have to add a line to the steering file.  We want to be able to turn this off to do studies of the two-hit pulses.

bloodyyugo (approved):
I'd prefer to see a switch in here that turns this on and off but since this change doesn't change any behavior in current code I guess it's ok.  We need to look into this more in the near future.

normangraf:
I agree that further study of the APV25 waveform fitting is merited. But I think the gains from this modification, especially since it does not change the results of our current reconstruction in any appreciable manner, warrant this update.

cbravo135:
It has not been demonstrated that there are any gains from this change. This only removes information so it is not possible anymore to further study the APV25 waveform fitting.

normangraf:
Will not pursue.

PR #876, opened 2021-06-09
Modify PhysicsRun2019MCRecon.lcsim to solve a bug 

### LINKED ISSUE #875: Fix bug in MC reconstruction steering files
A bug was introduced into the MC reconstruction steering files when the ReconParticleDriver_Kalman Driver was added. This resulted in the Ecal cluster corrections being applied twice. The following should resolve this issue:

<driver name="ReconParticleDriver_Kalman" type="org.hps.recon.particle.HpsReconParticleDriver" >
…
<applyClusterCorrections>false</applyClusterCorrections>
…

tongtongcao:
Modify PhysicsRun2019MCRecon.lcsim to solve a bug  that the Ecal cluster corrections are applied twice when recon particles by both GBL and KF tracks. Meanwhile, following what we did for 2016 MC, pedestals from DB are applied for MC, instead of running pedestal. So EcalRunningPedestal is removed.

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019MCRecon.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019MCRecon.lcsim
@@ -157,6 +157,8 @@
         <driver name="ReconParticleDriver" type="org.hps.recon.particle.HpsReconParticleDriver" > 
             <ecalClusterCollectionName>EcalClustersCorr</ecalClusterCollectionName>
             <trackCollectionNames>GBLTracks</trackCollectionNames>
+            <matcherTrackCollectionName>GBLTracks</matcherTrackCollectionName>
+            <trackClusterMatcherAlgo>TrackClusterMatcherMinDistance</trackClusterMatcherAlgo>
             <beamPositionX> 0 </beamPositionX>
             <beamSigmaX> 0.05 </beamSigmaX>
             <beamPositionY> 0 </beamPositionY>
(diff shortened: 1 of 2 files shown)


cbravo135 (changes requested):
The official recommendation for 2019 of the analysis working group is to use KF tracking only. https://confluence.slac.stanford.edu/display/hpsg/5.26.2021+Weekly?preview=/311544557/311546122/210526_kfSummary.pdf

Please change the order of ReconParticleDriver_Kalman and ReconParticleDriver and the configuration so the ecal cluster corrections are based on matching to the KF tracks. I do not mind to making this change for you if you want.

cbravo135 (approved):

tongtongcao:
> The official recommendation for 2019 of the analysis working group is to use KF tracking only. https://confluence.slac.stanford.edu/display/hpsg/5.26.2021+Weekly?preview=/311544557/311546122/210526_kfSummary.pdf
> 
> Please change the order of ReconParticleDriver_Kalman and ReconParticleDriver and the configuration so the ecal cluster corrections are based on matching to the KF tracks. I do not mind to making this change for you if you want.

I personally agree with you, Cameron. Please make the change as you suggested.

normangraf:
The steering file PhysicsRun2019MCRecon_LCIO.lcsim also needs to be modified.
Both steering files are also missing the OtherElectrons_KF collection name in the ReconParticleDriver_Kalman section.

cbravo135:
@normangraf Done

bloodyyugo (approved):

normangraf:
> @normangraf Done

Thank you.

PR #877, opened 2021-06-10
Iss851: Consolidate online monitoring modules

JeremyMcCormick:
The following modules were moved into the new online-monitoring module:

- ecal-event-display
- monitoring-util
- monitoring-app
- monitoring-drivers

This was done to reduce the overall build time, cut down on clutter at the top-level of the project, simplify dependencies, and put all code and resources under the org.hps.monitoring package in a single place. All files were moved using git so that their histories were left intact.

The online-recon module is kept as is because it has significantly different dependencies.


normangraf:
This looks like a very useful consolidation of these various modules. I don't feel qualified to rigorously test these changes, so I'll have to defer to @bloodyyugo and the subdetector experts who have run the online monitoring.

JeremyMcCormick:
The project builds successfully and the monitoring app starts fine.
Unlikely this broke anything since all the packages are the same, and
running from the generated assembly scripts or bin jar would not be
affected - that is all managed through distribution, which was not changed
aside from modifying the required deps to use the new module instead of the
old ones.

But more detailed verification can be done if necessary. Any bugs
introduced would anyways likely be quite minor.

normangraf (approved):
Consolidating these into one module should make things much easier to maintain.
All the tests run to completion.
Looks good.
Please proceed.

bloodyyugo (approved):

PR #881, opened 2021-07-19
Iss880

normangraf:
Tests all run successfully to completion. 
HelicalTrackHitRelations now correctly filled.


pbutti (approved):
I'm ok with this solution.

mholtrop (approved):
Simple solution for the problem.

PR #882, opened 2021-07-20
new track truth matching tool, and truth analysis driver

alspellm:
Adding tool for track->MCP truth matching, and an analysis driver to evaluate the matching. May or may not be useful


JeremyMcCormick:
I don't know if this is stale at this point but I don't see any reason this can't be merged since it is only adding additional analysis classes and not changing any existing ones.

cbravo135 (approved):
Safe addition of new analysis tools.

bloodyyugo (approved):

PR #885, opened 2021-08-23
Fix the Hodo monitoring

rafopar:
The Hodo hit collection is not 'SimpleGenericObject' anymore, but rather 'CalorimeterHit', The collection name is changed, and variable asignment satememts also changed.

### DIFF
--- a/online-monitoring/src/main/java/org/hps/monitoring/drivers/hodoscope/HodoscopePlots.java
+++ b/online-monitoring/src/main/java/org/hps/monitoring/drivers/hodoscope/HodoscopePlots.java
@@ -31,8 +31,8 @@ import org.lcsim.util.Driver;
 import org.lcsim.util.aida.AIDA;
 
 import org.lcsim.event.Cluster;
-
-import org.hps.recon.ecal.SimpleGenericObject;
+import org.lcsim.event.CalorimeterHit;
+//import org.hps.recon.ecal.SimpleGenericObject;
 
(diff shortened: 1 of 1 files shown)


rafopar:
It worked for me after those changes.

bloodyyugo:
It almost works!  For "matched hit energies" only one plot is filling and it has a lot of 0s for some reason...

rafopar:
Yes that is correct.

Others were not present event in the 2019 run.

It was more of diagnostic, just was testing something.
Later if I will have time, will popolate others as well, but it is not critical.

bloodyyugo (approved):
Looks good, thanks Rafo

normangraf:
Does this cause any problems of backwards-compatibility with the 2019 data?

rafopar:
No, I have tested this on 2019 data.

Before the reconstruction was making genericobjects for Hodo hits, now the recon produces Calorimeter hits, so accordingly I had to change the monitoring app to read from Calorimeter hits, instead of the genericObject.

bloodyyugo:
Probably for lcio files written before Maurik's changes (when?) but running from evio it's fine.

rafopar:
Yes, slcio file format got different.
I believe this change happened by the end of 2019, or early 2020.

Yes it is fine with EVIO.

normangraf (approved):
Looks good. Please proceed.

rafopar:
Thanks

PR #888, opened 2021-09-08
Iss883

pbutti:
- Added front-back fitting of tracks with KF
- Fixes to KF pattern reco

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/TrackUtils.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/TrackUtils.java
@@ -1934,9 +1934,10 @@ public class TrackUtils {
     
 
     //This method return the layer number and the module number of a sensor given the volume and the millepedeID 
-    public static org.hps.util.Pair<Integer,Integer> getLayerSide(int volume, int millepedeID) {
+    public static Pair<Integer,Integer> getLayerSide(int volume, int millepedeID) {
         
-        Integer retLy,retMod;
+        Integer retLy=null;
+        Integer retMod=null;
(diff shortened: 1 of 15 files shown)


JeremyMcCormick:
Just a quick tip based on scanning your code:

If you want your comments to show up as documentation in the javadoc attached to classes and methods then enclose them like this:

```
/**
  * This will show up in the javadoc!
  */
class MyClass {
    /**
      * So will this!
      */
      void something() {}
}
```

Double-slash comments will not:

```
// This will not show up in the javadoc. :(
void something() {}
```

Non-javadoc style is more appropriate for internal comments within methods.

Just mentioning this because you have some documentation-style comments on classes and methods but they will be stripped out if someone is looking at the online javadoc and won't be visible at all there.

JeremyMcCormick:
Also, a few more stylistic nitpicks:

We don't typically prepend underscores in Java code like this within this codebase:

```
_materialManager
```

Instead to clearly denote a class variable you can just use `this.materialManager` within the code. This will use the class variable even if `materialManager` was another variable declared in the method scope.

And here you may want to extract some of these hard-coded numbers out into constants:

```
Double chi1 = new Double(p1*t1.chi2s / t1.hits.size() + 300.*(1.0 - (double)t1.hits.size()/14.));
```

You can then change them easily in one spot and their meaning becomes much clearer to anyone reading the code.

bloodyyugo (approved):

normangraf (approved):
I've not done an in-depth review, but all tests pass.
Please proceed.

PR #892, opened 2021-09-10
Iss891

pbutti:
HotFix: Removed 2019 hardcoded survey constatns


normangraf:
How do these changes in the code affect the 2019 detectors? I would expect that if the hard-coded values were removed from the code that they would have to now appear in the compact.xml description of the 2019 detectors.

pbutti:
They break 2019 geometry. Yes, we can put these numbers as survey in the compact or as additional alignment corrections. They don't need to be hard coded in those classes

mholtrop:
Putting the numbers in the compact XML should be part of the merge, so that we don't break reprocessing of 2019 data. We're still far from done with 2019 :-(
I agree that these numbers should not have been hardcoded in the first place.

normangraf:
These changes to the code should remain on a dev branch until we can ensure backwards-compatibility with the 2019 detectors. If this really needs to get onto master right away then I would suggest renaming this version of HPSTracker2019GeometryDefinition.java to HPSTracker2021GeometryDefinition.java and modifying the 2021 compact.xml detector descriptions to instantiate that tracker.

pbutti:
The underlying structure of these classes is sub-optimal. Keep maintaining it is probably a wrong call. 
I'll do a dev-branch as you request. People that want to check 2021 should use then branch "Run2021" until this get resolved.

PR #897, opened 2022-02-08
Iss893: update Hodoscope reconstruction drivers to work for MC

tongtongcao:
Recently, it was found that hodoscope reconstruction drivers not to work for MC.
The drivers are developed to process data, and work well. However, there are several differences between data from Hodo evio reader and samples from MC reaout:
1) Output collection name is different
2) Cell ID is different
3) Detector element is not set for Hodo hits in MC readout, while it is set in Hodo evio reader.

The evio reader and MC readout were developed by different authors at different time. So format of output collections from them is not uniform. The issue has not been found before since we have not taken detailed analysis for MC recon Hodo samples.

With the updates, format of output collection from recon. drivers is uniform for data and MC. For ids of Hodo hits, the same logic is processed for MC as logic in Hodo evio reader. The logic is commented by Rafo in the Hodo evio reader:
        // ====================== Rafo =======================
        // Unlike to the ECal case, where each detector element is readout with a single channel
        // here for the hodoscope, there are detector elements (tiles) that are readout with two different channels
        // 
        // The code below, previously was assigning the hole value to 0, 
        // That way you would get an iD, which is ok for the detector element identification, howevere
        // This id is not correct, when one wants to knwo which PMT channel is actually this hit belongs to.
        // In particular during the recon, I was getting an error, that it can not find a cellID
        
        // Now as a possible solution (Possibly not the best solution), this method will return two IDs
        // The 1st one, is clauclulated taking into account the hit, while the 2nd one is calculated with 
        // the "hole=0" assumption.
        
        // Further depending on the intention, one can use ids.get(0) if one wants to check the hit ID, or ids.get(1), if onw
        // wants to identify the detector element.

Besides drivers, all related steering files for MC recon have been updated.

### DIFF
--- a/ecal-recon/src/main/java/org/hps/recon/ecal/HodoRawConverter.java
+++ b/ecal-recon/src/main/java/org/hps/recon/ecal/HodoRawConverter.java
@@ -2,12 +2,15 @@ package org.hps.recon.ecal;
 
 import org.lcsim.event.RawTrackerHit;
 import org.hps.conditions.hodoscope.HodoscopeConditions;
+import org.hps.conditions.hodoscope.HodoscopeChannel.GeometryId;
 import org.hps.conditions.hodoscope.HodoscopeChannel;
 import org.hps.conditions.hodoscope.HodoscopeChannelConstants;
+import org.lcsim.detector.identifier.IIdentifierHelper;
 import org.lcsim.event.CalorimeterHit;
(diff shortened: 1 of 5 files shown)


normangraf (approved):
Builds and all tests pass. Please proceed.

bloodyyugo (approved):

PR #906, opened 2022-04-18
Workaround for XXE injection vulnerability

cbravo135:
Dependabot alerted us of a vulnerability without a released fix yet:  https://github.com/JeffersonLab/hps-java/security/dependabot/3
This is the recommended workaround, we don't have to put this in because I guess we would take it out as soon as a fix is released. No idea of timeline for a jdom release with this fixed.


cbravo135:
Oops, I thought this only had one commit in it. Gonna remake this.

PR #907, opened 2022-04-18
Workaround for XXE vulnerability

cbravo135:
Dependabot alerted us of a vulnerability without a released fix yet: https://github.com/JeffersonLab/hps-java/security/dependabot/3
This is the recommended workaround, we don't have to put this in because I guess we would take it out as soon as a fix is released. No idea of timeline for a jdom release with this fixed.

### DIFF
--- a/conditions/src/main/java/org/hps/conditions/svt/SvtConfiguration.java
+++ b/conditions/src/main/java/org/hps/conditions/svt/SvtConfiguration.java
@@ -42,6 +42,8 @@ public final class SvtConfiguration extends BaseConditionsObject {
         final byte[] bytes = this.getFieldValue("content");
         final InputStream inputStream = new ByteArrayInputStream(bytes);
         final SAXBuilder builder = new SAXBuilder();
+        builder.setExpandEntities(false);
+
         builder.setValidation(false);
         return builder.build(inputStream);
     }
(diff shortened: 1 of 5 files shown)


baltzell:
Exactly what dependency is pulling this security issue in?  Do we need that dependency?

cbravo135:
@baltzell  As is explained in the dependabot alert linked above, JDOM. I am fairly certain we need this to parse all the xml file resources.

cbravo135:
You can see exactly where we instantiate SAXBuilder in this PR, it is exactly the line before the one I added to the files changed in this PR.

cbravo135:
Here is a link to more info on this vulnerability: https://github.com/advisories/GHSA-2363-cqg2-863c

JeremyMcCormick:
Cam, did you check that everything works properly after changing this?

We should double check that LCDD generates usable and correct output for SLIC and that the compact.xml is loaded correctly for recon.

If that's verified I don't see any issue with this update, especially since it would be easy to revert if we find later that it breaks something.

cbravo135:
I didn't test anything, just put in the suggested workaround. I will try to find some time to test this at some point if no one else has time until I can get to it.

cbravo135:
I just checked that the LCDD generated via org.hps.detector.DetectorConverter is identical before and after this change.

cbravo135:
Recon is able to load xml, doesn't crash. Both checks @JeremyMcCormick asked for have been done.

JeremyMcCormick (approved):
This looks reasonable to me.

tongtongcao (approved):
It makes sense for me to protect parser from XXE.

PR #908, opened 2022-04-19
Alignment updates

pbutti:


pbutti:
This PR modifies only files related to alignment and monitoring, with no impact on reconstruction chain. 
- I've added the support to pass the Energy from calorimeter clusters as momentum constraint, so the constraint can now be applied to all tracks and not only to FEEs.
- I've added the support to form a combined trajectory to align and at the same time reconstruct particle vertices <not tested and not used yet>. This implementation doesn't break anything as it is not used.
- I've made minor modifications to the monitoring plots for alignment.

cbravo135 (approved):
Thank you PF for making some nice additions to the alignment monitoring.

recon/src/main/java/org/hps/recon/vertexing/MultipleEventsVertexingDriver.java
```diff
@@ -105,12 +105,13 @@ protected void process(EventHeader event) {
                 //accumulatedTracks.add(track);
                 accumulatedBTracks.add(new BilliorTrack(track));
                 
-                if (track.getTrackStates().get(0).getTanLambda() > 0) {
+                //Avoid tracks too close to the beamline to have a more reliable result on z from single volumes
+                if (track.getTrackStates().get(0).getTanLambda() > 0.035) {
```
cbravo135, line 109: In general, I think these configuration parameters should be put into a steering file and not hard coded. Hard coding selections should be avoided in general.

bloodyyugo (approved):
This seems fine

recon/src/main/java/org/hps/recon/vertexing/MultipleEventsVertexingDriver.java
pbutti: Yeah. But this code is only for monitoring.

pbutti:
I have another iteration with the configurable selections. We should develop a trackSelector tool to keep those selections in a single place. At the moment they are copied everywhere multiple times.

PR #919, opened 2022-06-27
Fixing iss887

### LINKED ISSUE #887: ECAL calibrations for 2021

andrea-celentano:
ECAL 2021 energy calibration.


normangraf (approved):
Code builds cleanly and all tests pass.
FEE cluster energy peaks at beam energies.
Please proceed.

ecal-recon/src/main/java/org/hps/recon/ecal/IterateGainFactorDriver.java
```diff
@@ -52,16 +52,43 @@ public void setOutputCollectionName(final String outputCollectionName) {
...
+    
+    public void setCalibYear(int year) {
+        this.calibYear=year;
+        System.out.println("IterateGainFactorDriver: setting calib year to "+this.calibYear);  
+      
+    }
+    
+    
+    public void setBadChannels() {
+        if (this.calibYear==2021) {
```
cbravo135, line 78: I don't think we should be hard coding bad channel lists like this. The database is the preferred solution for these sorts of things.

ecal-recon/src/main/java/org/hps/recon/ecal/cluster/ClusterPositionCorrection2021.java
```diff
@@ -0,0 +1,258 @@
...
+
+    public double getY() {
+        return Y;
+    }
+
+}
+
+public final class ClusterPositionCorrection2021 {
+    // Parameterizations tested in MC using v3-fieldmap
+    // Nov 2015
```
cbravo135, line 52: We should use the database here rather than hard code numbers like this. Also, the code has a comment that these were tested on Nov 2015. I would like to see these tested them the 2021 beam energy and fieldmap.

cbravo135 (commented):
Let's see a talk on these changes at an Analysis/Reco meeting soon.

ecal-recon/src/main/java/org/hps/recon/ecal/IterateGainFactorDriver.java
baltzell: This is a calibration driver, not used at all in reconstruction/analysis, so not sure it's such a big deal.  There are gains in the database already which are appropriate for killing a channel in reconstruction and are already used for such.   And, since this is reading a gains file during calibrations, they can/should be killed from there too, so we can probably just delete or at least ignore these hard-coded numbers.

cbravo135:
https://indico.slac.stanford.edu/event/7201/contributions/3939/attachments/1667/4411/HPS-Ecal%20presentation%202021.pdf

cbravo135 (approved):

PR #922, opened 2022-08-02
add a steering file for overlaying MC signal with pulser data

tongtongcao:
A steering file for overlay MC signal with pulser data was developed.
Additionally, readout steering file with signal-pulserData as input was slight changed.

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/readout/Overlay.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/readout/Overlay.lcsim
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lcsim xmlns:xs="http://www.w3.org/2001/XMLSchema-instance" xs:noNamespaceSchemaLocation="http://www.lcsim.org/schemas/lcsim/1.0/lcsim.xsd">
+    <execute>
+        <driver name="OverlayDriver" />
+        <driver name="OutputDriver" />
+    </execute>
+    <drivers>
+        <driver name="OverlayDriver" type="org.hps.digi.DataOverlayDriver">
+            <!-- overlay file -->
+            <inputFile>${overlayFile}.slcio</inputFile>
+        </driver>
+        <driver name="OutputDriver" type="org.lcsim.util.loop.LCIODriver">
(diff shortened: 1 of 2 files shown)


cbravo135 (approved):
We should discuss what exactly we would like to see as validation for this. I am interested in seeing a comparison of our beam background MC and random triggers. Does this really only use pulser events from data or does it also use the faraday cup sourced random triggers? We should be able to use both, maybe do a quick comparison of the two to be sure they both appear to have zero bias. We might also want to look at some physics selections with beam background and compare to this, and see how close things look. This all being said, I am approving this because it is just adding some new steering files and not changing the production steering files.

tongtongcao:
> We should discuss what exactly we would like to see as validation for this. I am interested in seeing a comparison of our beam background MC and random triggers. Does this really only use pulser events from data or does it also use the faraday cup sourced random triggers? We should be able to use both, maybe do a quick comparison of the two to be sure they both appear to have zero bias. We might also want to look at some physics selections with beam background and compare to this, and see how close things look. This all being said, I am approving this because it is just adding some new steering files and not changing the production steering files.



> We should discuss what exactly we would like to see as validation for this. I am interested in seeing a comparison of our beam background MC and random triggers. Does this really only use pulser events from data or does it also use the faraday cup sourced random triggers? We should be able to use both, maybe do a quick comparison of the two to be sure they both appear to have zero bias. We might also want to look at some physics selections with beam background and compare to this, and see how close things look. This all being said, I am approving this because it is just adding some new steering files and not changing the production steering files.

Quickly answer the questions:
We did preliminary study to compare signal + pulser data to signal + MC beam, which were reported in several meetings. With loose kinematic cuts, they are basically consistent. One thing we are not clear is that MC beam has much higher FEE peak than pulser data.
For software itself, it does not care what random data as input. It is our choice to use what random data.
For the test study, random data with DAQ hps_v11_5_random_15kHz.cnf is applied.

JeremyMcCormick:
Noted Tongtong's answers to Cam's questions. This looks like steering files only and safe to merge.

JeremyMcCormick (approved):
Steering file only - looks safe.

PR #935, opened 2022-11-22
Add a steering file for monitoring using KF tracks and add a tool to remove hits from pattern reco in LCIO

pbutti:
Title says all.

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
@@ -1,6 +1,8 @@
 package org.hps.recon.tracking; 
 
 import java.util.List;
+import java.util.ArrayList;
+import java.util.Arrays;
 
 import org.lcsim.detector.tracker.silicon.SiSensor;
 import org.lcsim.event.EventHeader;
@@ -16,6 +18,8 @@ public class SensorSetup extends RawTrackerHitSensorSetup {
 
     /// Name of the collection of fitted hits 
     private String fittedHitColName_ = ""; 
+    
+    private List<String> sensorNames_ = new ArrayList<String>();
 
     /// Constructor
     public SensorSetup() { }
@@ -29,6 +33,10 @@ public class SensorSetup extends RawTrackerHitSensorSetup {
      */
     public void setFittedHitCollection(String fittedHitColName) { fittedHitColName_ = fittedHitColName; }
 
+    public void setSkipSensors(String[] sensorNames) {
+        this.sensorNames_ = new ArrayList<String>(Arrays.asList(sensorNames));
+    }
+    
     @Override
     protected void process(EventHeader event) { 
         
@@ -50,7 +58,26 @@ public class SensorSetup extends RawTrackerHitSensorSetup {
     public void loadFittedHits(List< LCRelation > fittedHits) {
 
         for (LCRelation fittedHit : fittedHits) { 
+
+            Boolean skipHit = false;
             RawTrackerHit rawHit = FittedRawTrackerHit.getRawTrackerHit(fittedHit);
+            
+            SiSensor sensor = (SiSensor) rawHit.getDetectorElement();
+            String name = sensor.getName();
+            
+            if (sensorNames_.size() > 0) {
+                
+                for (String sensorName : sensorNames_) {
+                    if (name.contains(sensorName)) {
+                        skipHit = true;
+                        break;
+                    }
+                }
+            }
+            
+            if (skipHit)
+                continue;
+            
             ((SiSensor) rawHit.getDetectorElement()).getReadout().addHit(fittedHit); 
         }
     }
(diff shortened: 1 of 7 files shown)


cbravo135 (approved):
Code added to SensorSetup should be safe since existing steering files won't have "SkipSensors" parameter list.

tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
```diff
@@ -50,7 +58,26 @@ protected void process(EventHeader event) {
...
 
         for (LCRelation fittedHit : fittedHits) { 
+
+            Boolean skipHit = false;
             RawTrackerHit rawHit = FittedRawTrackerHit.getRawTrackerHit(fittedHit);
+            
+            SiSensor sensor = (SiSensor) rawHit.getDetectorElement();
+            String name = sensor.getName();
+            
+            if (sensorNames_.size() > 0) {
```
omar-moreno, line 68: This check is unnecessary.  If sensorNames_ is empty, the whole for loop will be skipped.

tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
```diff
@@ -50,7 +58,26 @@ protected void process(EventHeader event) {
...
+
```
omar-moreno, line 69: I think the following works as well ... 
```
for (LCRelation fittedHit : fittedHits) { 

  RawTrackerHit rawHit = FittedRawTrackerHit.getRawTrackerHit(fittedHit);
  SiSensor sensor = (SiSensor) rawHit.getDetectorElement();
  if (sensorNames_.contains(sensor.name()) continue; 

  sensor.getReadout().addHit(fittedHit); 

}
```
Note the use of the `contains` method. When using a HashSet, this gives you O(1) lookup time vs O(n) when using an ArrayList or just iterating over the elements. HashSet is appropriate here since all of the sensor names will be unique. It also might be worth changing the name of the set  `sensorNames_` to `maskedSensorNames_` so it's clearer what it does.

tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
```diff
@@ -16,6 +18,8 @@ public class SensorSetup extends RawTrackerHitSensorSetup {
 
     /// Name of the collection of fitted hits 
     private String fittedHitColName_ = ""; 
+    
+    private List<String> sensorNames_ = new ArrayList<String>();
```
omar-moreno, line 22: Change this to a HashSet. You don't need to initialize the array since you do so in the `setSkipSensors` method below.

tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
```diff
@@ -29,6 +33,10 @@ public SensorSetup() { }
      */
     public void setFittedHitCollection(String fittedHitColName) { fittedHitColName_ = fittedHitColName; }
 
+    public void setSkipSensors(String[] sensorNames) {
+        this.sensorNames_ = new ArrayList<String>(Arrays.asList(sensorNames));
```
omar-moreno, line 37: Change this to Hash Map. 

On a separate note, I wonder if this method should check if `sensorNames` was already initializing before doing it again.  I can see a situation where this gets set twice on accident in a steering file and could lead to unexpected behavior. Maybe if `sensorNames` has already been initialized, throw an exception?

omar-moreno (changes requested):

tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
cbravo135: The behavior of this will be different. contains behavior is different when called on an array vs a string. I think we do indeed want contains called on the string and not the array, which gives us easier control over which sensors we drop. This is only called once, so the complexity of the code doesn't really matter, and either way will effectively run in the same time. We would like to be able to remove multiple sensors with a single string if they all contain the same sub-string.

tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
cbravo135: I think the framework already checks you only set a specific variable in the steering file once and throws an error if it is there more than once.

cbravo135: Given the behavior we would like from this code, not clear to me HashMap really buys us anything.

tracking/src/main/java/org/hps/recon/tracking/SensorSetup.java
cbravo135: The loop is over all the hits, this check is on the list of sensors we want to skip. I do think sensorNames_ is potentially a confusing variable name since it is really the sensors to be skipped.

omar-moreno (approved):
I don't agree with these changes but will approve it anyway. The performance issues can be taken care of in the future.

PR #939, opened 2022-12-03
Add tanLambda cuts

sarahgaiser:
work in progress

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/gbl/GBLOutputDriver.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/gbl/GBLOutputDriver.java
@@ -80,6 +80,9 @@ public class GBLOutputDriver extends Driver {
 
     private double minPhi = -999.9;
     private double maxPhi = 999.9;
+
+    private double minTanL = 0.015;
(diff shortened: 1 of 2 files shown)


pbutti:
Hi Sarah, the cuts should be done at track reconstruction level instead at plotting level. 
Making cuts at track reconstruction level (i.e. SimpleGBLTrajAliDriver) would harmonize what is used for alignment and what is used for monitoring. 
In fact the tanLambda cuts can already be applied at that stage see:
https://github.com/JeffersonLab/hps-java/blob/master/tracking/src/main/java/org/hps/recon/tracking/gbl/SimpleGBLTrajAliDriver.java#L352

cbravo135:
Hi PF, there are times when I personally run GBLOutputDriver and not SimpleGBLTrajAliDriver, and would like to be able to have these kinds of cuts possible there too.

pbutti:
Is it clear why CI fails?

cbravo135:
Relaunched CI and they passed. @pbutti

cbravo135 (approved):

pbutti (approved):

PR #940, opened 2022-12-06
Avoid order O(clusters.size()) operation when debug is disabled.

omar-moreno:


cbravo135 (approved):
Thanks for the hotfix

alspellm (approved):
bien

PR #941, opened 2022-12-08
Add a standalone driver to make final state particles.

### LINKED ISSUE #942: ReconParticleDriver --> FinalStateParticleDriver and VertexingDriver
Currently, ReconParticleDriver handles

- The creation of track-cluster pairs via the specified matcher.
- The creation of ReconstructedParticles from track-cluster matches and from unmatched tracks and clusters.
- Vertexing of the resulting ReconstructedParticles using several algorithms.
- The creation of "V0" particles from the resulting vertexed particles. 

The driver has become difficult to work with so this issue aims to breakup the driver into simpler, easier to work with drivers. This will be achieved by doing the following
- Adding a driver used to create FinalStateParticles only. For now, the driver will also handle calling of the track-cluster matcher.
- Adding a driver that uses the collection of FinalStateParticles to creat "v0" particles. Instead of having a single driver use all fitting algorithms, only a single algorithm will be used per driver.  The use of different fitting algos can be achieved by using a different instance per algo. 

To-Do: 

- [x] Creation of FinalStateParticles (#941)
- [ ] Electron and positron collections are created in FinalStateParticleDriver and marked transient.
- [ ] Creation of V0 particles.
- [ ] Validate and update steering files

Notes: 
- Currently, the method TrackClusterMatcher::matchTracksToClusters requires a List<List<Track>> as a argument.  This should be update such that the method only requires a List<Track> and an additional method that handles a List<List<Track>> can be added.
- The method TrackClusterMatcher::matchTracksToClusters should be simplified. A lot of these parameters should be set independently or passed via the constructor. 
- Are the "MOUSE" cuts being used? If not, the use of the class should be removed.

omar-moreno:
This is part of a series of changes aimed at simplifying HpsReconParticleDriver.

### DIFF
--- a/recon/src/main/java/org/hps/recon/particle/FinalStateParticleDriver.java
+++ b/recon/src/main/java/org/hps/recon/particle/FinalStateParticleDriver.java
@@ -1,0 +1,364 @@
+package org.hps.recon.particle;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+
+import org.lcsim.event.base.BaseReconstructedParticle;
+import org.lcsim.event.base.BaseCluster;
+import org.lcsim.event.Cluster;
+import org.lcsim.event.EventHeader;
+import org.lcsim.event.ReconstructedParticle;
+import org.lcsim.event.Track;
+import org.lcsim.geometry.Detector;
+import org.lcsim.geometry.subdetector.HPSEcal3;
+import org.lcsim.util.Driver; 
+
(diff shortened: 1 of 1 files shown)


cbravo135 (approved):
Single commit adding a Driver which is not yet instantiated in any steering file, so generally safe. Thanks Omar!

sarahgaiser:
I only understand half of it but looks great in my opinion. I appreciate the comments explaining everything!

sarahgaiser (approved):

PR #943, opened 2023-01-14
Create the electron and positron FinalStateParticle collections

omar-moreno:


cbravo135:
Thanks for adding this. I think it would be nice to be able to choose the collection names in the steering file, and use this as also a way to configure which particular collections will be put into the event by not including ones that have "" as the destination collection name. This will just make this more flexible, and allow us to better control the size of the output reconstruction files in the types of jobs that typically include this driver.

PR #945, opened 2023-02-10
remove odd-number-hit cut for Kalman tracks

### LINKED ISSUE #944: Known Inconsistencies with KF-based Alignment
There are two (small) inconsistencies left on `master` that I think can be removed now that we are more confident in using KF-based Alignment.

- [ ] Removal of odd-number-hit tracks. This was originally done to make the comparison to ST-based Alignment more simple, but we should not be throwing away e.g. 13-hit tracks because that is the whole reason we wanted to switch to KF-based Alignment.

https://github.com/JeffersonLab/hps-java/blob/2216d8cdc97aab9cd3b83acbcbf6ac4923553ad2/tracking/src/main/java/org/hps/recon/tracking/gbl/SimpleGBLTrajAliDriver.java#L577-L584

- [ ] The number-of-hits cut in the alignment driver is using the old ST-based style of counting hits (i.e. a module where both sensors combine to form a single hit). We should abandon this style in favor of the simpler KF-based style of counting (each sensor hit is a single hit).
    - I'm not fully committed to doing this since it would then require all steering files that use this driver to be updated.

https://github.com/JeffersonLab/hps-java/blob/2216d8cdc97aab9cd3b83acbcbf6ac4923553ad2/tracking/src/main/java/org/hps/recon/tracking/gbl/SimpleGBLTrajAliDriver.java#L556-L560

tomeichlersmith:
This was a relic of a time when we were comparing KF-based and ST-based alignments. It is not needed anymore and we should remove this cut so we can use all tracks found by KF, including ones that may have only one sensor from a given module.

## Still To Do
- [ ] Check that this does not break previous alignment studies
- [ ] Maybe see if it helps?
- [ ] Maybe change the meaning of the nhit cut as described in #944

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/gbl/SimpleGBLTrajAliDriver.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/gbl/SimpleGBLTrajAliDriver.java
@@ -573,14 +573,6 @@ public class SimpleGBLTrajAliDriver extends Driver {
                         || (tanLambda < 0 && track.getTrackerHits().size() < actualHitCut))  {
                     continue;
                 }
-
-                if (TrackType == 1 && track.getTrackerHits().size() % 2 == 1) {
-                    // this is a KF track with an odd number of hits which /cannot/
-                    // be equivalent to a GBL track so we are going to skip it
-                    // in a future where KF-based alignment is the standard,
-                    // we probably want to remove this
-                    continue;
-                }
                 
                 // ask tracks only on a side
                 if (trackSide >= 0) {


pbutti:
@tomeichlersmith Status of this? Still WIP ?

tomeichlersmith:
I just haven't tested it - it compiles and runs normally if you want to just risk it.

pbutti:
ok. LGTM

pbutti (approved):

sarahgaiser (approved):

PR #954, opened 2023-03-18
Iss953

normangraf:
Adding an energy threshold to the MC processing commensurate with that used in the hardware during the 2019 and 2021 data-taking runs.


mholtrop (approved):
Minimal changes.

andrea-celentano (approved):

PR #956, opened 2023-03-20
Reduce log level of SVT filter messages.

JeremyMcCormick:
Change the log level of the SVT filter messages from WARN to FINE to reduce console output when running with the default log settings.


normangraf (approved):
It would be useful to accumulate and print out a summary of how many events failed each of the flags in the endOfData method.

pbutti (approved):

JeremyMcCormick:
I'm going to merge this to cut down on unnecessary test output. I'll implement printouts of totals in another PR.

PR #970, opened 2023-04-05
Fix number of DOF in KF tracks

### LINKED ISSUE #969: Consistency fix between NDF and TrackerHits in KF tracks
@robertprestonjohnson Noticed that there is an inconsistency between the tracker hits and the number of DoF stored in the KF tracks. 
The reason I think is line 
https://github.com/JeffersonLab/hps-java/blob/7e9a7eac08b6229d4f8b8bf02e29cbcc432215b4/tracking/src/main/java/org/hps/recon/tracking/kalman/KalmanInterface.java#L598

where we should pass the number of tracker hits (that get checked by ID > 0) instead of the size of the measurement sites vector.

pbutti:
fixes #969

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/kalman/KalmanInterface.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/kalman/KalmanInterface.java
@@ -595,7 +595,7 @@ public class KalmanInterface {
         
         // other track properties
         newTrack.setChisq(kT.chi2);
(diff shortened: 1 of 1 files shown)


normangraf (commented):
Fixed:

image

normangraf (approved):
Would have been tested sooner if git branch had been eponymous with the git issue.

cbravo135 (approved):

PR #975, opened 2023-04-05
Add support for Github Actions plus cleanup and updates to POM files

JeremyMcCormick:
This PR implements Github Actions for hps-java CI/CD.

The new workflows are as follows:

- `deploy` - Deploys the snapshot (development) jars whenever the master branch is updated
- `release` - Performs the Maven release, creates the Github release page, (optionally) updates the project website, and bumps the development version
- `site` - Deploys the project website to gh-pages
- `test` - Runs the tests for PRs (replaces the old `maven.yml` workflow file)

All of these can be run manually through the "Actions" tab in the project.

Additionally, I made a cleanup pass on all the project POM files and some other project configuration, including the following:

- Updated all Maven plugins to their current release versions
- Changed deployment so it only uploads the bin jar rather than the jars for every module (which never made much sense as they are not really used as independent artifacts by other projects)
- Moved all module dependency versions into the `dependencyManagement` section of the main POM
- Removed outdated or unused configuration and plugins from the main POM
- Updated SCM settings for usage on Github
- Removed unwanted dependency on jdom 1.0 which was conflicting with 1.1
- Simplified and cleaned up the project website
- Removed old collaborator contact information 
- Added a log4j config to remove warning messages from the tests
 
My TODO list before un-drafting:

- [x] Replace "JeremyMcCormick" user with "JeffersonLab" group in all workflows and POM configuration
- [x] Add a deployment key and its private key as a secret
- [x] Add the CI deployment username and password secrets
- [x] Change the development release to match hps-java/master branch (should be 5.2.0-SNAPSHOT)
- [x] Confirm that running the release workflow with tests works on the fork without errors
- [x] Confirm that the site workflow works correctly on the branch and that the site navigation works correctly in gh-pages

After being approved, this should be merged to master as a "squash commit" to avoid unwanted messages from the branch history (from merge commits, test releases, etc.).


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,28 @@
+name: deploy
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+    # Ignore release tags
+    tags-ignore:
+      - hps-java-*
```
tomeichlersmith, line 9: Maybe I'm confused on what "deploy" means, but wouldn't we want maven to deploy _only_ on releases instead of everything else?

site/site.xml
```diff
@@ -0,0 +1,49 @@
...
+    <skin>
+        <groupId>org.apache.maven.skins</groupId>
+        <artifactId>maven-fluido-skin</artifactId>
+        <version>1.11.1</version>
+    </skin>
+
+    <bannerRight>
+        <name>HPS</name>
+        <src>https://confluence.slac.stanford.edu/download/attachments/84248493/heavy_photon_logo_50.jpg</src>
+        <href>https://jeremymccormick.github.io/hps-java/</href>
```
tomeichlersmith: Missed a link update

```suggestion
        <href>https://jeffersonlab.github.io/hps-java/</href>
```

.github/workflows/deploy.yml
JeremyMcCormick: > Maybe I'm confused on what "deploy" means, but wouldn't we want maven to deploy _only_ on releases instead of everything else?

It is Maven terminology - a deployment means uploading the snapshot (development) version of the jar whenever a merge is made to the master branch. This makes the current development version of the bin jar available for download from the Nexus repository.

Release jars are also uploaded this way but in a different workflow.

tomeichlersmith (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,166 @@
...
+    secrets:
+      SSH_PRIVATE_KEY:
+        required: true
+      CI_DEPLOY_USERNAME:
+        required: true
+      CI_DEPLOY_PASSWORD:
+        required: true
+
+jobs:
+  # TODO: Add test here for a sanity check
```
sarahgaiser, line 59: Is this still an active todo?

sarahgaiser (approved):

JeremyMcCormick:
I can do this later in another PR.

PR #996, opened 2023-06-08
Add track state at target

alspellm:
Extrap track to target and make new track state where track params ref = (trackX,trackY,0), therefore d0 = 0. 
trackY position may replace d0 if using this track state in analysis. 

Track reference point is set to (trackX, trackY, trackZ), so track position at target is saved.

May want to change track param reference point to ref = (trackX,0,0), so that d0 != 0.

These track states only exist if target position is set in steering-file KalmanPatRecDriver.


tracking/src/main/java/org/hps/recon/tracking/TrackUtils.java
```diff
@@ -755,6 +755,57 @@ public static BaseTrackState getTrackExtrapAtHodoRK(Track trk, FieldMap fM, int
...
+        double[] params = getParametersFromPointAndMomentum(finalPos, momTrans, (int) charge, bFieldY);
+        BaseTrackState bts = new BaseTrackState(params, bFieldY);
+        //reference point is set to track position in X Y Z
+        bts.setReferencePoint(finalPos.v());
+        //Consider shifting reference point to (trackX, 0, 0)...otherwise d0=0
+        boolean case2 = false;
+        if (case2){
+            double [] paramsRefPoint = {finalPos.x(), finalPos.y(), 0.0};
+            bts.setReferencePoint(paramsRefPoint);
+            double[] newRef = {-4.3, 0.0, 0.0};
```
cbravo135: I think you want to change -4.3 to target_z

alspellm:
Ah you are correct, whoops

bloodyyugo (approved):

normangraf:
Instead of having magic code triggered by the addition of the targetPosition variable please add a boolean which explicitly states what is going to happen followed by the target location, such as

``` <createTrackStateAtTargetPosition>true</createTrackStateAtTargetPosition> ```
``` <targetPosition>-4.3</targetPosition> ```

pbutti (approved):

PR #1001, opened 2023-07-19
remove wrong shim thickness and typo for top back UChannel

sarahgaiser:
Removes 700um shift that Cam notices during his Survey Constant analysis and fixes bug in geometry code. The bug was as follows: by not setting the node for the top back alignment corrections, the corresponding survey volume had no access to the survey constants defined in the `compact.xml`. Fixing the typo solves this problem.


cbravo135 (approved):
I checked out this code and made comparisons between the 2016 and 2019 detectors and confirmed this does indeed fix the issue of getting different detectors with the same survey data.

detector-model/src/main/java/org/lcsim/geometry/compact/converter/HPSTracker2019GeometryDefinition.java
```diff
@@ -556,7 +556,7 @@ public static class ModuleL3Bot extends ModuleL13Bot {
         //2019 MRSolt survey
         //protected final static double L3_new_vertical_shift = 0.7 - 0.012;
         //2021 nominal
-        protected final static double L3_new_vertical_shift = 0.7;
+        protected final static double L3_new_vertical_shift = 0.;
```
pbutti, line 559: Any clue why Matt Added this 0.7mm. ?

detector-model/src/main/java/org/lcsim/geometry/compact/converter/HPSTracker2019GeometryDefinition.java
```diff
@@ -51,7 +51,7 @@ public void build() {
         AlignmentCorrection supBotCorrBack = getUChannelCorrection(false,90);
         supBotCorrBack.setNode(node);
         AlignmentCorrection supTopCorrBack = getUChannelCorrection(true,90);
-        supTopCorr.setNode(node);
+        supTopCorrBack.setNode(node);
```
pbutti, line 54: Indeed a bug!

pbutti (approved):

detector-model/src/main/java/org/lcsim/geometry/compact/converter/HPSTracker2019GeometryDefinition.java
cbravo135: My best guess is this was to account for the shims added wrt the engineering run detector. I don't think it was that thick, will confirm with Tim. Also, this isn't even used if you supply survey data for the modules.

cbravo135 (changes requested):
Need to properly update detector lcdd files.

normangraf:
Please hold off on merging this into master until we can discuss this at the software meeting. I would much prefer that we maintain backwards-compatibility with four years of simulation and reconstruction work. Changing the geometry in this way will invalidate those detectors. We should modify the compact.xml files to be commensurate with the new code such that the output geometry is unchanged. We should then proceed from here on out with new detectors that correct the mistake.

cbravo135 (approved):
We have not done much simulation/reconstruction work in the past 4 years. If we need to rereco some MC for some reason we can always use the 5.1 release to have the same detectors. Making things "backwards compatible" to be able to keep a bug around is poor software practice.

PR #1003, opened 2023-07-27
restore relevant detectors to old version

sarahgaiser:
tl;dr:
DETNAME_fixed: always use this version
DETNAME: only use this version if you need pre-bug-fix (wrong) geometry

To ensure that the same detector name always refers to the same geometry, I went back and adjusted the detectors pointed out by @normangraf to represent the same geometry as before the geometry bug fix (https://github.com/JeffersonLab/hps-java/pull/1001). This includes removing the top back survey constants (which have previously not been used due to a typo) and introducing a 700um shift of layers 3 and 4 towards the beam. I copied the unmodified version to DETNAME_fixed which should be used from now on for further alignment and analysis work unless you want to specifically analyze MC data generated before the fix or other scenarios where you need the pre-bug-fix version of the detector.

### DIFF
--- a/detector-data/detectors/HPS_PhysicsRun2019_Pass2/HPS_PhysicsRun2019_Pass2.lcdd
+++ b/detector-data/detectors/HPS_PhysicsRun2019_Pass2/HPS_PhysicsRun2019_Pass2.lcdd
@@ -2,7 +2,7 @@
 <lcdd xmlns:lcdd="http://www.lcsim.org/schemas/lcdd/1.0" xmlns:xs="http://www.w3.org/2001/XMLSchema-instance" xs:noNamespaceSchemaLocation="http://www.lcsim.org/schemas/lcdd/1.0/lcdd.xsd">
   <header>
     <detector name="HPS_PhysicsRun2019_Pass2" />
-    <generator name="lcsim" version="1.0" file="/sdf/group/hps/users/sgaiser/src/hps-java/detector-data/detectors/HPS_PhysicsRun2019_Pass2/compact.xml" checksum="3526012259" />
+    <generator name="lcsim" version="1.0" file="/sdf/group/hps/users/sgaiser/src/hps-java/detector-data/detectors/HPS_PhysicsRun2019_Pass2/compact.xml" checksum="1490788426" />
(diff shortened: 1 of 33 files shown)


sarahgaiser:
Before merging this we should check that dumping the pre-bug-fix geometry and the adjusted geometry from the PR give the same result.

normangraf:
Thanks for doing this. I don't see any real need for the *_fixed detectors since we should be proceeding with new and improved detectors for pass1. But I have no objections if others might find them useful.

sarahgaiser:
I just added the *_fixed detectors to keep track of the "original" version of those detectors in case we need them again. Did you happen to check if dumping the old and fixed geometries yields the same results?

normangraf:
I have not checked the geometries.

cbravo135:
There are some samples made with HPS-PhysicsRun2019-v2-4pt5 too
https://confluence.slac.stanford.edu/display/hpsg/SLIC+level+for+pass0+of+2019+MC
Also HPS_PhysicsRun2019-v2-FEE-Pass0
https://confluence.slac.stanford.edu/display/hpsg/FEE+sample+for+comparison+of+GTP+clusters+between+readout+and+hardware
Also HPS-v2019-3pt7GeV
https://confluence.slac.stanford.edu/display/hpsg/SLIC+level+for+pass0b+of+future+3.7+GeV
HPS-v2019-2pt3GeV  ----> https://confluence.slac.stanford.edu/display/hpsg/SLIC+level+for+pass0+of+future+2.3+GeV
HPS-v2019-1pt056GeV ---> https://confluence.slac.stanford.edu/display/hpsg/SLIC+level+for+pass0b+of+future+1.056+GeV
HPS-v2021-1pt92GeV ----> https://confluence.slac.stanford.edu/display/hpsg/SLIC+level+for+pass0+of+future+1.92+GeV

tongtongcao:
To not change existing scripts for MC/data production, I suggest to directly update the original versions without changing detector names, and copy original versions with other names, like DETNAME_original.

pbutti:
@sarahgaiser 
What's the status with this PR ?

PR #1004, opened 2023-08-30
Pre bug hps_run2021pass1_v3

sarahgaiser:
Restore HPS_Run2021Pass1_v3 detector to give same results as pre-bugfix


cbravo135 (approved):
Thanks!

PR #1019, opened 2023-11-08
Update readout system with DAQ configuration for 2016 MC

tongtongcao:
Here is a talk for description of updates.
[Apply DAQ Configuration into Software Readout System for 2016 MC.pdf](https://github.com/JeffersonLab/hps-java/files/13301311/Apply.DAQ.Configuration.into.Software.Readout.System.for.2016.MC.pdf)
Due to the updates, track-cluster time offset in reconstruction steering files changes (8 ns shift).


record-util/src/main/java/org/hps/record/daqconfig/DAQConfigDriver.java
```diff
@@ -252,4 +330,29 @@ public void setDataFileRepository(String filepath) {
...
+    }
+    
+    /**
+     * According to run number, a specified DAQ version name is returned.
+     * @param runNumber
+     * @return name of a DAQ configuration version
+     */
+    private String mapBetweenRunNumberDAQVersion(int runNumber) {
+        
+        // 2019 experiment
```
cbravo135: this should be 2016

cbravo135 (approved):

bloodyyugo (approved):

PR #1024, opened 2024-02-13
Change setup for trackClusterTimeOffset and beam sigma in 2016 MC reconstruction steering files

tongtongcao:
The timing of the ECal changes when using the daq config for the trigger instead of the old way, this is a change that should have happened before when these files were added. We are also updating the beam profile parameters as extracted by the analysis of Alic.


cbravo135 (approved):
Thanks TongTong!

sarahgaiser (changes requested):
Can you add a short description of why you are changing the setup and a few words about the new values?

tongtongcao:
> Can you add a short description of why you are changing the setup and a few words about the new values?

Since software trigger system was updated with DAQ configuration, time offset between track and cluster is changed. According to Alic's test, the shift comparing to the old time offset is 8 ns (i.e., 2 clock cycles). 
Besides, beam size is reset based on Alic's study.

sarahgaiser (approved):

PR #1030, opened 2024-04-30
Move the track extrapolation from dipole edge to z=800

### LINKED ISSUE #1026: Track extrapolation to ECAL fix
As documented by Alic, Norman, and Lewis the track state at the ECAL is not computed properly. 
This issue is to fix the problem.

pbutti:
This should fix #1026 
@mholtrop

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/BeamlineConstants.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/BeamlineConstants.java
@@ -18,6 +18,8 @@ public final class BeamlineConstants {
 
     public static final double DIPOLE_EDGE_ENG_RUN = 457.2 + 1080 / 2;
 
+    public static final double ECAL_TRACK_EXTRAPOLATION_START_Z = 800; // Starting point for track extrapolation to ECAL -- MWH 2024/4/15
+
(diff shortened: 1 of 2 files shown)


bloodyyugo (changes requested):
One of the tests is failing

mholtrop:
> One of the tests is failing

Error:    PhysRun2016V0ReconTest>ReconTest.testRecon:380->ReconTest.comparePlots:252->ReconTest.compareTrees:272->ReconTest.compareHistograms:293 Bin 12 is different for: V0 Bottom Track Momentum expected:<241> but was:<240>

So that is because our tests are not very smart and do not permit improvements of the code.

pbutti:
@mholtrop 
Can we change the reference with the new file?

mholtrop:
>  Can we change the reference with the new file?

We should be able to, I think. I know how though.

pbutti:
This PR is linked to #1031

cbravo135:
Just want to point out that there isn't a need to request changes when a test is failing since that already blocks the merge anyway.

cbravo135:
I updated the reference file, the tests are all passing for this branch now.

cbravo135 (approved):
Nice clean way to do this.

bloodyyugo (approved):

PR #1039, opened 2024-05-29
adding overlay for 2016

sarahgaiser:


tongtongcao:
Thanks for the nice work.

Here are my comments for coding:
Basically, the updates are for application of 2016 DAQ configuration in the new digitization system.
So Similar updates can be taken as PR#1019 for the updates for application of 2016 DAQ configuration.

1) For EcalDigitizationWithPulserDataMergingReadout2016Driver.java, do not need to add this new driver. Instead, you can update EcalDigitizationWithPulserDataMergingReadoutDriver.java. Refer to [driver](https://github.com/JeffersonLab/hps-java/blob/master/ecal-readout-sim/src/main/java/org/hps/readout/ecal/updated/EcalDigitizationReadoutDriver.java) for how to do updates.

2) For the new steering file, since DAQ configuration is applied, parameters for digitization and trigger system are loaded from the DAQ configuration and they are run dependent, so do not set them in the steering file. Refer to [steering file](https://github.com/JeffersonLab/hps-java/blob/master/steering-files/src/main/resources/org/hps/steering/readout/PhysicsRun2016TrigPairs.lcsim)

tongtongcao:
Looks fine for me. But validation is needed before approval. 
Keep one thing in mind:
According to experience of 2019&2021, peak for delta_time between track and cluster might change due to the new readout system. So \<trackClusterTimeOffset\>36.8\</trackClusterTimeOffset\> in reconstruction steering files might need to be updated.

bloodyyugo:
Is this still relevant?

PR #1049, opened 2024-09-18
Type change from soft link to actual file to fix resource loading iss…

### LINKED ISSUE #1048: Resource files are not properly loaded in ClusterEnergyCorrection2021 or related classes.
Java will get an incorrect InputStream that does not actually read the contents of the file. This error is quite Java version and architecture dependent.
Issue is caused by the resource files being soft links rather than a file, which leads to the unpredictable behavior.

mholtrop:
Simply fix: change soft linked files to the actual files.


sarahgaiser:
Hi Maurik, could you explain the issue again, briefly? I think I missed most of your explanation during the meeting...
I saw the description of iss_1048 this PR is fixing but could you explain again how you encountered this problem?

mholtrop:
@sarahgaiser The issue is that if you have soft links (i.e. files that refer to other files, which you create with "ln -s old_name new_name") in your Java resources, then the reading of these files is not well defined, i.e. all bets are off when it will work and when not. It turns out that so far it has worked on Linux with the versions of Java we used, but the newer version of Java on my Mac, it does not work as expected. 
The correct way to solve this is to not have links but make actual copies of the files (i.e. "cp old_name new_name"). There were only a few files that were links, and so I copied them instead of linked them.
I hope that explains it.

sarahgaiser:
Note: I checked building this on my M1 mac and it worked fine.

bloodyyugo (approved):

sarahgaiser (approved):

PR #1062, opened 2024-12-09
Fix an issue in SinglesTrigger2019ReadoutDriver 

tongtongcao:
The driver is supposed to process any single trigger.
However, codes https://github.com/JeffersonLab/hps-java/blob/054ee5d43d5e36618d248d41a6b8c124d2be3304/ecal-readout-sim/src/main/java/org/hps/readout/trigger2019/SinglesTrigger2019ReadoutDriver.java#L145C9-L155C27 requires that sizes for both GTP clusters and hodo patterns are larger than 0.
It is requirement for single3 trigger. Update for the codes was missed when I did readout update for processing multiple triggers.
The bug causes issue when we process other single triggers except single3 trigger.


bloodyyugo (approved):
Looks good

sarahgaiser (approved):

PR #1064, opened 2024-12-10
add readout steering file for 2021 MC with single2&single3 triggers

tongtongcao:
The new steering file for 2021 MC with single2&single3 triggers is a copy of the steering file for 2019 MC: PhysicsRun2019TrigSinglesWithPulserDataMerging.lcsim.
The only change is factorGainConversion from 0.000833333 for 4.55 GeV beam of 2019 to 0.0008125 for 3.74 GeV beam of 2021. 
The factor is for hodoscope gain conversion from self-defined unit/ADC to MeV/ADC. 
Parameter setup is determined by distribution of energy deposit for hodoscope hits from SLIC.
Effect of the change on hodoscope patterns in the readout system is almost negligible since the change is slight and threshold for hodoscope hits is quite low.


bloodyyugo (approved):

mgignac (approved):
Looks good to me -- only (picky) comment would be whether you want to be explicit and specify the `triggerType` in the SinglesTrigger3 instead of rely on the default in the source code, incase e.g. someone were to change the source code at another time. Anyways, approved.

PR #1067, opened 2024-12-15
Update PhysicsRun2021_pass0_recon.lcsim

rodwyer100:
I am adding validated kalman filter tracking algorithm parameter optimizations to the steering file for the first pass. The optimization method you are familiar with (likely from the last collaboration meeting) was used w/ 1000 iterations to obtain settings which improved MC efficiency 4% and increased data track number 10% in run 14185. I believe the reason MC only increased marginally compared to data track number is because MC already is quite clean and easy to construct well. The info pertaining to MC optimization (and its validation) was already included in previous talks. I had to manually include new optimization constraints for data, because I realized that some parameters relevant to data could not have been optimized for MC. The reward function was updated to optimized efficiency in MC and track no in a single data file in run 14185; the MC efficiency is not increasing as high as it did before after 1000 iterations but I suspect as much is expected because now it has to simultaneously work in a clean and messy environment. I am doing some further validation, but I would be comfortable using this steering file in the run. Here is a plot showing the track number increase between the two iterations:

image
Red is old, blue is new; it is shifted into higher track no per event. In total we find 10% more tracks while retaining the same MC efficiency and negligible change in fake rate or duplicates (atleast when training on 14185). I can characterize its behavior on other runs, and more extensively if requested.

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass0_recon.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass0_recon.lcsim
@@ -98,10 +98,36 @@
 	    <debug>false</debug>
         </driver>
         <driver name="KalmanPatRecDriver" type="org.hps.recon.tracking.kalman.KalmanPatRecDriver">
-            <!--<doDebugPlots>false</doDebugPlots>-->
-            <!-- <siHitsLimit>50</siHitsLimit> -->
-            <seedCompThr>0.05</seedCompThr>
-            <verbose>false</verbose>
+	    <numPatRecIteration> 3 </numPatRecIteration>
+            <numKalmanIteration> 1 </numKalmanIteration>
+            <maxPtInverse> 8.757651 </maxPtInverse>
+            <maxDZero> 38.0487 </maxDZero>
+            <maxZZero> 3.98915 </maxZZero>
+            <maxChiTwo> 11.777395 </maxChiTwo>
+            <minHits> 0  </minHits>
+            <minStereo> 3  </minStereo>
+            <maxSharedHits> 3 </maxSharedHits>
+            <maxTimeRange> 39.95028 </maxTimeRange>
+            <maxTanLambda> 8.186345 </maxTanLambda>
+            <maxResidual> 13.71568 </maxResidual>
+            <maxChiTwoInc> 13.52662 </maxChiTwoInc>
+            <minChiTwoIncBad> 7.00678 </minChiTwoIncBad>
+            <maxResidShare> 13.967129 </maxResidShare>
+            <maxChiTwoIncShare> 9.771546584 </maxChiTwoIncShare>
+            <mxChiTwoVtx> 1.7652935 </mxChiTwoVtx>
+            <numEvtPlots> 5 </numEvtPlots>
+            <doDebugPlots> false </doDebugPlots>
+            <siHitsLimit> 466 </siHitsLimit>
+            <seedCompThr> .725912 </seedCompThr>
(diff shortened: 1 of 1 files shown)


steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass0_recon.lcsim
```diff
@@ -98,10 +98,36 @@
...
         <driver name="KalmanPatRecDriver" type="org.hps.recon.tracking.kalman.KalmanPatRecDriver">
-            <!--<doDebugPlots>false</doDebugPlots>-->
-            <!-- <siHitsLimit>50</siHitsLimit> -->
-            <seedCompThr>0.05</seedCompThr>
-            <verbose>false</verbose>
+	    <numPatRecIteration> 3 </numPatRecIteration>
+            <numKalmanIteration> 1 </numKalmanIteration>
+            <maxPtInverse> 8.757651 </maxPtInverse>
+            <maxDZero> 38.0487 </maxDZero>
+            <maxZZero> 3.98915 </maxZZero>
```
mgignac, line 105: Is this a cut on the z0 impact parameter? What is z0 referenced to? (0,0,0) or the beam-spot? If the latter, this cut may be problematic run-by-run, since the beam-spot may move around run-to-run, and we are assuming a fixed (0,0,0) position at the moment. From what I've seen in my alignment jobs, there is considerable variations throughout the dataset. It may be advisable to loose this cut further or remove completely until we have a better understanding / calibrated the beam-spot run-by-run. What do you think?

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass0_recon.lcsim
```diff
@@ -98,10 +98,36 @@
...
+            <maxChiTwo> 11.777395 </maxChiTwo>
+            <minHits> 0  </minHits>
+            <minStereo> 3  </minStereo>
+            <maxSharedHits> 3 </maxSharedHits>
+            <maxTimeRange> 39.95028 </maxTimeRange>
```
mgignac, line 110: What time range does this correspond to?

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass0_recon.lcsim
mgignac: I went back and checked the plots I showed in the collaboration meeting (see slide 35): https://indico.slac.stanford.edu/event/9401/contributions/10967/attachments/4845/13012/2024_11_21_alignment_collaboration-meeting.pdf 

Even if we have variation, it's much much smaller than the cuts values you have here. So I don't think this is a concern. It should also be checked for displaced signals, but we can save that for later.

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass0_recon.lcsim
mgignac: As discussed in the meeting, this cut should be loose enough.

mgignac (approved):

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass0_recon.lcsim
bloodyyugo: Very picky comment but can we change the these to:  
      <maxPtInverse> 8.75 </maxPtInverse>
      <maxDZero> 38.0 </maxDZero>
      <maxZZero> 4.0 </maxZZero>
      
      etc....so that there are reasonable sig figs in these numbers.

bloodyyugo (approved):

PR #1072, opened 2024-12-20
build for a newer java version

baltzell:


mgignac:
Hi @baltzell -- thanks for working on this! From what I can tell, the CI builds in JDK 21 were successful. Is this simple to deploy to JLab and start some testing? It would be great to have this for the next 10% production (~1-2 month timescale from now). Maybe Zhaozhong could run O(10) runs from the 1% pass and we can validate this?

baltzell:
Let's ask Zhaozhong for 2 tests, one with the main branch and java 21 at runtime (by setting JAVA_HOME), and then again with this branch (built for and requires 21).  I think I posted on Slack a while back, /work/hallb/hps/baltzell/hps-java-21 was this branch in December, I could get it up to date with the main branch ...

bloodyyugo:
So, has Zhaozhong done the "(built for and requires 21)" test yet?  I think not...but maybe I'm confused.

PR #1075, opened 2025-01-08
Iss1050:  svt hit position and time smearing for MC

### LINKED ISSUE #1050: Add driver that performs SVT cluster position smearing
Add a driver to be used (optionally) to smear the local SVT cluster positions in MC.   

The goal is to better duplicate the hit residuals and track chi2 seen data but the plan is to start simple and add complexity. 

- step 0:  single smearing factor (or absolute number) for all clusters over all layers and nHits
- step 1:   separate factors for 1 & 2 hit clusters
- step 2:  also separate by layers in some way (eg L0+L1 vs the rest or L0+L1, L2+L3, outer layers)
- step 3:  position dependent smearing ... we would need some guidance from data for this

I don't plan to include everything in this issue, probably just up to step 2.

bloodyyugo:
This adds classes and example smearing files for smearing of svt cluster positions and fitted raw hit t0s. The values are roughly tuned on 2016 data and will need to be redone for each data set.  

I note that while the t0 smearing makes the MC hit time resolutions roughly match data, the track time still seems too narrow.  

This pull request closes #1050


bloodyyugo:
I should mention this pull also includes changes to the KFOutputDriver in particular removing the 3d histograms.

mgignac (approved):

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2016FullReconMC_KF_Killer_Smearer.lcsim
```diff
@@ -0,0 +1,203 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lcsim xmlns:xs="http://www.w3.org/2001/XMLSchema-instance" xs:noNamespaceSchemaLocation="http://www.lcsim.org/schemas/lcsim/1.0/lcsim.xsd">
+    <!-- 
+      @brief Steering file that will be used for pass 1 reconstruction of
```
sarahgaiser: Can we update this description?

sarahgaiser: Or just add a comment here why we are adding this steering-file for 2016 now

bloodyyugo: Changed comment

sarahgaiser (approved):

PR #1077, opened 2025-01-22
Adding SVT hit pattern

mgignac:
In an effort to drop the hit collections, we need to add information about hits on layers for downstream analysis. An existing method in LC-SIM called `setSubdetectorHitNumbers` seems to be what we need, but previously isn't filled. In this PR, this information is added.


bloodyyugo (approved):
Seems fine...

sarahgaiser (approved):

PR #1078, opened 2025-01-27
Recon file for pass 1 2021 data 

mgignac:
Identical to the pass0 recon file, but with additional collections (mainly hit collections) removed.


bloodyyugo (approved):
I think this looks good.

mholtrop (approved):
I agree, this all looks fine.

PR #1081, opened 2025-02-13
comment out the maxResidual setting in KalmanPatRecDriver block

bloodyyugo:
It was the cause of the problems with the momentum...I think it was basically kicking out good hits.  The default for this is 50.0, the setter in the steering file had it at ~13, which is apparently too low.  

See [https://jira.slac.stanford.edu/browse/HPSANA-62](https://jira.slac.stanford.edu/browse/HPSANA-62)

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass1_recon.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass1_recon.lcsim
@@ -109,7 +109,7 @@
             <maxSharedHits> 3 </maxSharedHits>
             <maxTimeRange> 39.95028 </maxTimeRange>
             <maxTanLambda> 8.186345 </maxTanLambda>
-            <maxResidual> 13.71568 </maxResidual>
+<!--            <maxResidual> 13.71568 </maxResidual>   -->
             <maxChi2Inc> 13.52662 </maxChi2Inc>
             <minChi2IncBad> 7.00678 </minChi2IncBad>
             <maxResidShare> 13.967129 </maxResidShare>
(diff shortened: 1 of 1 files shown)


steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass1_recon.lcsim
```diff
@@ -109,7 +109,7 @@
             <maxSharedHits> 3 </maxSharedHits>
             <maxTimeRange> 39.95028 </maxTimeRange>
             <maxTanLambda> 8.186345 </maxTanLambda>
-            <maxResidual> 13.71568 </maxResidual>
+<!--            <maxResidual> 13.71568 </maxResidual>   -->
             <maxChi2Inc> 13.52662 </maxChi2Inc>
             <minChi2IncBad> 7.00678 </minChi2IncBad>
             <maxResidShare> 13.967129 </maxResidShare>
```
mgignac: I wonder about this parameter too? I guess the maximum residual for shared hits? Probably not as pronounced as the `maxResidual`, but maybe it causes some other issues too? What is the default here?

bloodyyugo: Good point...It can't have too big effect from what I see, but I'll comment it out too just to be consistent.

mgignac (approved):

sarahgaiser (approved):

sarahgaiser:
We will have to do the same for the MC drivers, don't we?

mgignac:
> We will have to do the same for the MC drivers, don't we?

Yes, we also need to update the MC steering file as well (then re-run MC)

PR #1086, opened 2025-03-04
Classes for skimming recon events and a possible V0 skim setup

### LINKED ISSUE #1085: Add capability to skim events to different streams & add potential v0 skim code for 10%

bloodyyugo:
closes issue #1085


mgignac (approved):
Code looks fine to me, thanks! Of course this will need to be tested. One thing we may need to confirm is whether the cuts are valid across all runs, but that can be confirmed later.

baltzell (approved):

PR #1089, opened 2025-03-31
Inverse Variance Track Time Averages.

rodwyer100:
It is mathematically proven that the way to take a weighted average of independently, possibly nonidentically distributed, normal distributions is through inverse variance sampling. Namely you weight a time sample by the inverse of the square of its error. This push request aims to implement this change. It does so fairly successfully; I include a number of html files with various information about how the changes work. There are two caveats about this weighting scheme that are important to note. They first arises from the broad shoulder issue we have in the first two layers; the second from just hps-java architecture. That is, due to an irreducible degeneracy in hits (that was only partially solved by previous chi square probability manipulation) there is a significant degree of non gaussianity to the first two layer hit time distributions (particularly in layers nearmost to the detector). 
1. This seems to occasionally make the inverse variance timing average worse. To counteract this, the first two layers are removed from the averaging step. Don't worry; this doesn't mean they aren't involved in momentum and other things they are critical for; their removal just noticeably and significantly improves the mean. This makes this change almost uniformly an improvement.
2. The HPS track time is calculated from Measurement site object in java. This is an object that, while it retains amplitude error information, it forgets everything to do with the time error. You can be sure about this because the KalmanPlotting classes have to re-obtain the Hit objects whenever they want to plot information about the time errors of hits. To circumvent this issue, I rerun fitting for hits on track to obtain their hit time error at the time that we cant to calculate track time. Luckily this change is ~10 lines; we don't need to do all the time translation junk that has to be done at the hit time step because the time error is invariant under time translation. Thats with the exception of the pileUp decision, where we retain the time closer to the track time. In this ase, I make the reasonable assumption that the hit time error of the chosen hit will be less than the other and use that time error (this is almost always true). It is with this choice I see the almost uniform improvement in the following plots:

Track Time:https://s3df.slac.stanford.edu/people/rodwyer1/timeFix/orChange/tracktime.html
Z Momentum of Track:https://s3df.slac.stanford.edu/people/rodwyer1/timeFix/orChange/pzonTrack.html (seemingly completely unchanged).

The change is small, but theoretically motivated. When yall look at the plots, and are convinced it is fine, I will also make changes to the jlab plot stuff (that will probably need to be done again)

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
@@ -19,6 +19,8 @@ import org.ejml.dense.row.factory.LinearSolverFactory_DDRM;
 import org.ejml.interfaces.linsol.LinearSolverDense;
 import org.hps.util.Pair;
 
+import org.lcsim.event.TrackerHit;
+
 /**
  * Track followed and fitted by the Kalman filter
  */
@@ -138,6 +140,7 @@ public class KalTrack {
         Cp = null;
         // Fill the maps
         time = 0.;
+	double terr = 0.0;
         tMin = 9.9e9;
         tMax = -9.9e9;
         this.chi2 = 0.;
@@ -150,7 +153,9 @@ public class KalTrack {
                 continue;
             }
             nHits++;
-            time += site.m.hits.get(site.hitID).time;
+	    double hTV = site.m.hits.get(site.hitID).timeErr;
+            time += site.m.hits.get(site.hitID).time/hTV;
+	    terr += 1.0/hTV;
             tMin = Math.min(tMin, site.m.hits.get(site.hitID).time);
             tMax = Math.max(tMax, site.m.hits.get(site.hitID).time);
             this.chi2 += site.chi2inc;
@@ -158,7 +163,7 @@ public class KalTrack {
                 System.out.format("  Layer %d, chi^2 increment=%10.5f, a=%s\n", site.m.Layer, site.chi2inc, site.aS.helix.a.toString());
             }
         }
-        time = time / (double) nHits;
+        time = time / terr;
         reducedChi2 = chi2 / (double) nHits;
         lyrMap = null;
         millipedeMap = null;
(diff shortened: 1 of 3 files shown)


rodwyer100:
Btw for those html plots, the changes were done on the distributions on the right (unchanged left). Only the first is slight worse; all other runs are slightly better.

tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
```diff
@@ -150,15 +153,17 @@ public class KalTrack {
                 continue;
             }
             nHits++;
-            time += site.m.hits.get(site.hitID).time;
+	    double hTV = site.m.hits.get(site.hitID).timeErr;
+            time += site.m.hits.get(site.hitID).time/hTV;
+	    terr += 1.0/hTV;
```
mgignac, line 158: Can you explain & confirm why you are taking the inverse of the timeErr here? From below, I see you already accounted for the inverse sum of the errors by inverting the Variance after the for loop, so wanted to confirm this isn't being accidentally done twice.

mgignac: Never mind -- I also missed the /hTV on L157. This all looks fine. Thanks!

rodwyer100: So the time average is a weighted sum. Its sum c_{i}t_{i}/sum c_{i}. In this case, c_{i} is 1/htV. Hence you add to the time time/htV (where htV is the time Error) and you add 1/hTV to some constant I am calling terr. At the end, below this, I take time = time/terr, which gives you the weighted sum. The only thing to check I think is that timeErr is truly the variance, but it is to my understanding.

mgignac (approved):

mgignac (approved):

tracking/src/main/java/org/hps/recon/tracking/kalman/KalmanInterface.java
```diff
@@ -1058,7 +1077,24 @@ private boolean fillAllMeasurements(EventHeader event) {
...
+                                if(fit.getT0Err()<Min){
+                                        Min=fit.getT0Err();
+                                }
+                        }
+                        if(module.Layer>1){
+                                Variance+=1/(Min*Min);
+                        }
+                }
+                Variance=1.0/Variance;
+                Measurement m = new Measurement(umeas, xStrip, du, time, localHit.getdEdx()*1000000.,Variance);
```
bloodyyugo, line 1097: Really, this should be done in clustering but SiTrackerHit doesn't have a place for the error on the time.  We do extend SiTrackerHit in hps-java at org.hps.recon.tracking.SiTrackerHitStrip1D and we could put the error in there but it wouldn't be saved to the LCIO.  For now this is ok...I think we have some other changes to LCIO/lcsim we need to make so we should put this on the list.

bloodyyugo (approved):

PR #1113, opened 2025-06-27
Recon steering file for hit killing/smearing studies

mgignac:
Add kinks/residuals
Save all collections
Update vertex positions (not important for this study)

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass3_recon_hitstudies.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass3_recon_hitstudies.lcsim
@@ -158,11 +158,11 @@
             <includeUnmatchedTracksInFSP>true</includeUnmatchedTracksInFSP>
             <useInternalVertexXYPositions>false</useInternalVertexXYPositions>
             <requireClustersForV0>false</requireClustersForV0>
-            <beamPositionX>0.180</beamPositionX>
+            <beamPositionX>0.100</beamPositionX>
             <beamSigmaX>0.350</beamSigmaX>
-            <beamPositionY>0.100</beamPositionY>
+            <beamPositionY>0.06</beamPositionY>
             <beamSigmaY>0.100</beamSigmaY>
-            <beamPositionZ>0.900</beamPositionZ>
+            <beamPositionZ>-0.5</beamPositionZ>
             <maxElectronP>7.0</maxElectronP>
             <maxVertexP>7.0</maxVertexP>
             <minVertexChisqProb>0.0</minVertexChisqProb>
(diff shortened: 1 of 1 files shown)


bloodyyugo:
This will only save only events passing the V0 skim cuts.  I think this is probably good but I don't think it's how I did this in the past.  I added Elizabeth to the reviewers list...comments?

mgignac:
> This will only save only events passing the V0 skim cuts. I think this is probably good but I don't think it's how I did this in the past. I added Elizabeth to the reviewers list...comments?

Ah, for the purpose of validating the skimming, we should save both. How is that toggled?

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass3_recon_hitstudies.lcsim
```diff
@@ -0,0 +1,237 @@
...
+	    <makeMollerCols>true</makeMollerCols>
+            <unconstrainedMollerCandidatesColName>UnconstrainedMollerCandidates_KF</unconstrainedMollerCandidatesColName>
+            <unconstrainedMollerVerticesColName>UnconstrainedMollerVertices_KF</unconstrainedMollerVerticesColName>
+            <beamConMollerCandidatesColName>BeamspotConstrainedMollerCandidates_KF</beamConMollerCandidatesColName>
+            <beamConMollerVerticesColName>BeamspotConstrainedMollerVertices_KF</beamConMollerVerticesColName>
+            <targetConMollerCandidatesColName>TargetConstrainedMollerCandidates_KF</targetConMollerCandidatesColName>
+            <targetConMollerVerticesColName>TargetConstrainedMollerVertices_KF</targetConMollerVerticesColName>
+        </driver>  
+        <driver name="StripEvent" type="org.hps.recon.skims.MultiSkimDriver">
+	    <skimV0>true</skimV0>
```
EBerzin, line 185: I think the V0 skimming is toggled here. We didn't have this in previous versions of the steering files that were used for the hit killing/smearing studies, but it should be okay if we want to leave it (validation of the MC-data comparison will require some basic selections anyway).

mgignac: Thanks. Indeed -- however, for the purpose of this small pass, I'd also like to validate the skimming closes, so I'd like to save both skimmed and unskimmed outputs. @bloodyyugo was able to do this in the past, not sure which flag allows for that.

baltzell (approved):

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2021_pass3_recon_hitstudies.lcsim
mgignac: I just tested this steering file -- I do see two slcio outputs, one v0skim and the nominal output of all events.

bloodyyugo: This is because this is also doing the LCIOWriter at the end.  The skimmer is writing the streams while LCIO writer is writing every event.

bloodyyugo (approved):

Issue #1134, opened 2025-07-25
use SLAC nexus server for hps-lcim deployment and download


bloodyyugo:
We now have access to publish the hps-lcsim packages to the SLAC SRS nexus server.  This issue is to read them from there, just a couple of changes to the pom.xml

bloodyyugo:
closed with PR 1135

PR #1135, opened 2025-07-25
change pom settings to pull from slac nexus

bloodyyugo:
closes issue #1134


mgignac (approved):
Awesome, thanks for following up with this!

mholtrop (approved):
Works. This is much better than setting tokens in GitHub.

PR #1142, opened 2025-09-05
Adding preliminary hit smearing and hit killing

EBerzin:
This includes hit efficiency corrections derived from two runs, and the corresponding steering files: `PhysicsRun2021MCRecon_KF_WithSpacing_pass3_HitKill_14232.lcsim`, `PhysicsRun2021MCRecon_KF_WithSpacing_pass3_HitKill_14464.lcsim`.

`PhysicsRun2021MCRecon_KF_WithSpacing_pass3_HitSmear.lcsim` implements the full smearing chain, including position, hit time, and track time smearing. This required adding a new `TrackTimeSmearer` driver, as well as modifications to the existing smearers to ignore data hits.


sarahgaiser:
Here are the changes that need to applied to the recon steering files to get the correct MC track and cluster times.
The `EcalTimeCorrection` driver is already part of the steering files, but has to be configured a bit differently.
```
<driver name="EcalTimeCorrection" type="org.hps.recon.ecal.EcalTimeCorrectionDriver">
        <additionalShift>13.3</additionalShift>
</driver>
```
In `RawTrackerHitFitterDriver` set `tsCorrectionScale` to 165 (should be 120 before this change)
```
<tsCorrectionScale>165</tsCorrectionScale>
```
This PR should be merged after https://github.com/JeffersonLab/hps-java/pull/1143 which includes all other necessary changes.

mgignac:
For the next MC production, I assume we will want to use:
`PhysicsRun2021MCRecon_KF_WithSpacing_pass3_HitSmear.lcsim` 

The PR looks good to me, but we will need to add the lines Sarah pointed out above to the above steering file before we can proceed. When you get a chance, can you make those changes? Would be great if we could get this running before the weekend, if possible...

sarahgaiser (approved):
Looks good to me!

mgignac (approved):
Looks good to me too, thanks!

PR #1148, opened 2025-09-23
Iss1147 -- Add hole-accounting to tracks

### LINKED ISSUE #1147: add track holes to TrackData
Add list of track holes to track data.  A hole is where a reconstructed track's trajectory is predicted to go through the active area of a sensor but no hit is assigned to the track.

bloodyyugo:
Closes #1147 

Add hole on track--defined as the layer where the track is projected to go through the sensitive area of a sensor but no hit is included in the track.  

This is included in the Track object's _subdetId (see here [Track](https://github.com/JeffersonLab/hps-lcsim/blob/391d46492257c5ab155e22ecf74734efd7aaa27a/event-model/src/main/java/org/lcsim/event/base/BaseTrack.java#L386-L388)).  We already use this to account for the layer numbers with hits on track, setting _subdetId[layer]=1 for a  hit on track.  For a hole, I've set _subdetId[layer]=-1.

### DIFF
--- a/tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
+++ b/tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
@@ -295,27 +295,49 @@ public class KalTrack {
             Vec intPnt = HelixState.atPhi(site.aS.helix.X0, aStar, phiInt, site.aS.helix.alpha);
             // Transform the intersection back into the global coordinates
             Vec globalInt = site.aS.helix.toGlobal(intPnt);
+	    //	    System.out.println("KalTrack::unbiasedIntersect  global x = "+globalInt.v[0]+
+	    //	       "  y = "+globalInt.v[1]+
+	    //		       "  z = "+globalInt.v[2]);
+	    
             if (!local) {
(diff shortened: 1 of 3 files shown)


mgignac (approved):
Looks good to me, thanks!

Only comment: I'm not sure if it's the GitHub viewer, but it I'd be nice to intend your code  after the definition of a method. Would be a lot easier to read.

bloodyyugo:
That is weird, because in emacs these are indented correctly.  


> Looks good to me, thanks!
> 
> Only comment: I'm not sure if it's the GitHub viewer, but it I'd be nice to intend your code after the definition of a method. Would be a lot easier to read.

mholtrop:
> That is weird, because in emacs these are indented correctly.
> 
> > Looks good to me, thanks!
> > Only comment: I'm not sure if it's the GitHub viewer, but it I'd be nice to intend your code after the definition of a method. Would be a lot easier to read.

Maybe run "untabify" in Emacs to turn tabs to spaces?

bloodyyugo:
That was it, thanks!


> > That is weird, because in emacs these are indented correctly.
> > > Looks good to me, thanks!
> > > Only comment: I'm not sure if it's the GitHub viewer, but it I'd be nice to intend your code after the definition of a method. Would be a lot easier to read.
> 
> Maybe run "untabify" in Emacs to turn tabs to spaces?

Issue #1149, opened 2025-09-30
add separate nHits requirements for top & bottom and for each track finding iteration in KF


bloodyyugo:
We currently perform track finding separately for top and bottom SVT halves and do two iterations of track finding, the second with looser requirements than the first.  However, there is only a single "minHits" setter in KalmanPatRecDriver...it uses the minHits to set the first iteration to minHits+1 and second to minhits and the top and bottom has same requirement.  

We'd like to be able to set the minHits separately for each iteration and also for top & bottom (particularly since 2021 is missing a bottom module-5 sensor).

PR #1150, opened 2025-09-30
Add more flexible setting for minHits into KF track finding

bloodyyugo:
This closes issue #1149


rodwyer100:
This looks like a pretty standard change; I wonder if you retained all the plots in the TruthTrack processor, but the logic looks sound (and I am already running it so I know it isn't breaking things for sure).

rodwyer100 (approved):

mgignac (approved):
Looks good to me... approving! 

My only comment, which doesn't need to hold this up, is whether the configuration level changes should go into a `pass4` steering file, so we preserve the behavior of our last pass and mark these changes for the upcoming pass?

Issue #1152, opened 2025-10-14
Use the supplied beamspot to set the target position when making AtTarget track state


bloodyyugo:
Use the supplied beamspot to set the target position when making AtTarget track state

PR #1153, opened 2025-10-14
Fix target plane position for AtTarget TrackState

bloodyyugo:
Closes issue #1152


rodwyer100:
I don't see this causing a problem algorithmically (its clean code), and atleast conceptually we said we want to use the magnetic field in the center of the SVT our value (and this is closer). Have you seen whether this change does something large? If it doesn't I think I could just say a blank yes on this.

mgignac (approved):

rodwyer100 (approved):

PR #1159, opened 2025-12-04
Add isolations to TrackData object for Kalman tracks

### LINKED ISSUE #1158: Calculate and include hit-on-track isolations for KF tracks
Hit isolations were never included when we switched over to Kalman Tracking.

bloodyyugo:
Isolations were never calculated/added to the TrackData for Kalman tracking.  This PR does that.  

-- For each layer, find the closest Si cluster, in the direction away from the beam, to the cluster-on-track and calculate the distance between the two.  
    --  if there is no cluster found or no hit-on-track for that layer, returns -999.
--  include this list of layer-isolations in the TrackData (which is a GenericObject) as the double array (which is it's intended purpose already).  

The hpstr code already uses this array, in utils::buildTrack, to store the isolations in the Track object so no need to modify that code.  

Closes #1158


mgignac (commented):
Thanks @bloodyyugo !

Looks good to me. In the hpstr level code, there is a timing cut to ensure the close by hit is in time (<30ns). Do we want to introduce a similar cut? You retrieve the time of the closest hit, but as far as I understand this information get dropped when written to the TrackData? There is also the possibility that there are >1 hits with slightly more spatial separation but closer in time, I suppose. It's probably a small effect. Perhaps a close approximation would be to write isot0 hit times as well?

tracking/src/main/java/org/hps/recon/tracking/kalman/KalmanPatRecDriver.java
```diff
@@ -537,13 +517,7 @@ private ArrayList<KalTrack>[] prepareTrackCollections(EventHeader event, List<Tr
                 //Add the TrackResiduals
                 List<Integer> layers    = new ArrayList<Integer>();
 		List<Double> residuals  = new ArrayList<Double>();
-                List<Float> sigmas      = new ArrayList<Float>(); 
-		List<Integer> layersInt = new ArrayList<Integer>();
```
sarahgaiser, line 541: Is there a reason why this chunk of code is moved down a few lines? It is hard to parse what is going on with the different loops and if statements.

tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
```diff
@@ -514,6 +515,49 @@ public Pair<Double[], Double> unbiasedIntersect(int layer, boolean local) {
...
+             return new Pair<Double, Double>(-999., -999.);
+         }         
+    }
+
+    public Pair<Double, Double> getIsoAndT0(MeasurementSite ms){
+        double iso=-999.;
+        double isot0=-999.;
+        SiModule sensor=ms.m; 
+        List<Measurement> allHits=sensor.hits;
+        if(allHits.size()>2 && ms.hitID>-1){
```
sarahgaiser: why is this >2? is it because you need at least two hits to form a (strip?)cluster for that sensor?

tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
```diff
@@ -514,6 +515,49 @@ public Pair<Double[], Double> unbiasedIntersect(int layer, boolean local) {
...
+            Measurement hitOnTrack= allHits.get(ms.hitID);
+            double[] hpsSensorPos=KalmanInterface.localKalToHps(sensor.p.X());
```
sarahgaiser: is hpsSensorPos supposed to be used again?

tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
```diff
@@ -514,6 +515,49 @@ public Pair<Double[], Double> unbiasedIntersect(int layer, boolean local) {
...
+            // keep track of sensor position & orientation
+            // use the sign of the global plane position (vertical = z)
+            // times the local-->global v-->z element
+            // to determine if sensor is aligned (+ive v --> away from beam)
+            // or anti-aligned (+ive v is towards beam)
```
sarahgaiser, line 541: I don't fully understand this comment. What do you mean by `local-->global v-->z element` and what is `+ive v`

tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
```diff
@@ -514,6 +515,49 @@ public Pair<Double[], Double> unbiasedIntersect(int layer, boolean local) {
...
+
+            double vertPos=sensor.p.X().v[2];
+            double l2gv2z=sensor.R.M[1][2];
+            int awayFromBeam = (int)Math.signum(vertPos*l2gv2z);
+            //sort the hits on the module by increasing v
+            Collections.sort(allHits,Measurement.MeasurementComparatorUp);
+            // now get the position of nearest hit _away_ from beam
+            // the hits on the SiModule are sorted by v
+            int isoID=ms.hitID+awayFromBeam*1;
```
sarahgaiser: is the `*1` here needed? because either this is supposed to be `*(-1)` and meaningful or it does nothing.

sarahgaiser:
I think this looks good now. @bloodyyugo could you address my two questions above? happy to approve the PR after you have answered them!

tracking/src/main/java/org/hps/recon/tracking/kalman/KalmanPatRecDriver.java
bloodyyugo: The first loop is of hits on track (so, only layers with a HOT are included) and the second one is over all layers.  I separated the intersections list instantiations down to just before the loop over all layers because that's where it's filled.

tracking/src/main/java/org/hps/recon/tracking/kalman/KalTrack.java
bloodyyugo: The local-->global refers to the rotation matrix R (local i.e. sensor frame to the "global" frame for kalman tracking) and v-->z element is the element in that matrix (the z component of the v direction).  +ive == positive.

sarahgaiser (approved):

mgignac (approved):

PR #1160, opened 2025-12-15
Adding moller and physics recon steering files

mgignac:
Clone of pass5 steering file with modifications:

- Target position is -1.1mm (-0.75mm) for Physics (Moller) runs
- Skimming outputs only; Mollers are skimmed with simple two electron selection; physics is usual v0 skimming.


sarahgaiser (approved):
looks good to me

rodwyer100 (approved):
These seem to be the steering files I am familiar with, should work fine

bloodyyugo (approved):

PR #1161, opened 2026-01-03
working version of MC steering

sarahgaiser:
I had some issues with the first version of this PR, so I'm opening a new one. This should now only contain the updated steering file for pass5 physics MC recon.


mgignac (approved):
Looks good!

bloodyyugo (approved):

rodwyer100 (approved):
In so far as it relates to things I've optimized, looks like a sensible file!

PR #1162, opened 2026-01-13
Adding pass5_v9 smearing and killing constants, hit-based smearing

EBerzin:
Hit smearing and preliminary hit killing based on the pass5_v9 run. 

Right now things are organized with a separate steering file for each run number we plan to use, but I'm happy to reformat that if we want. The difference is only in the hit-on-track efficiencies. I've only included constants for runs 14272 and 14349 so far.


bloodyyugo (approved):

mgignac:
Thanks @EBerzin ! It looks good, but just a question about the two run numbers: as far as I understand, deriving both the smearing and hit killing will require MC generated with detector conditions matching that run number. Besides using the correct detector, a major difference will be the y-offset will change between 14272 and 14369. Was this accounted for ?

EBerzin:
It wasn't...I used only MC generated with detector 14272 to derive all of the corrections. So maybe it makes sense to wait until we have MC with the correct offsets available and re-calculate the efficiency corrections then?

mgignac:
OK, thanks -- yes, I think it makes sense to derive these corrections as "official" MC arrives for the various runs periods. Then we know they are in sync. So perhaps remove the 14369 steering file for now, then this is ready to go?

sarahgaiser (approved):

mgignac (changes requested):
Should we remove the 14369 efficiency corrections as well, just so we don't confuse ourselves once we circle back to that?

mgignac (approved):

PR #1163, opened 2026-02-19
adding steering file for recon of late pulser runs

sarahgaiser:
For later pulser runs there is a 0.12mm shift in beam Y which has to be added to the recon steering file.
I also renamed the "default" steering file for earlier runs for more clarity.


mgignac (approved):
Looks good, thanks!

mholtrop (approved):
Adding a file. No reason not to approve.

PR #1165, opened 2026-03-09
Fix issue with the RfFitterDriver crashing if the RF pulse is absent.

mholtrop:
The code should not crash if the RF pulse is not detected. We have some early runs in 2021 where the RF pulse was not plugged in, so the data is garbage. We should still be able to analyze such data without needing to change the steering file.

The fix is simple, make a common sense guard against buffer over run (if you access i+1 do not loop to size()), and test if a peak was actually found before using the results.


mholtrop:
Note I also remove two comments that were left in ThreeBodySkimmer.java and FEESkimmer.java.

baltzell (approved):

bloodyyugo (approved):

PR #1166, opened 2026-03-18
Adding steering files for 2019 pass2

EBerzin:
Added ability to skip sensor in RawTrackerHitFitterDriver.

Added steering files with new seeding strategies, masking L5b.

### DIFF
--- a/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass2_recon.lcsim
+++ b/steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass2_recon.lcsim
@@ -1,0 +1,261 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<lcsim xmlns:xs="http://www.w3.org/2001/XMLSchema-instance" xs:noNamespaceSchemaLocation="http://www.lcsim.org/schemas/lcsim/1.0/lcsim.xsd">
+    <!-- 
+      Steering file for pass-2 2019 reconstruction on readout EVIO data 
+      created:  03/17/2026
+      @author Elizabeth Berzin <eberzin@slac.stanford.edu>
+    -->
+    <execute>
+
+        <driver name="EventMarkerDriver"/>
+        <driver name="EventFlagFilter"/>
+
+        <!--RF driver-->	
+        <driver name="RfFitter"/>	
+
+        <!-- Ecal reconstruction drivers -->        
+        <driver name="EcalRunningPedestal"/> 
+        <driver name="EcalRawConverter" />
+        <driver name="EcalTimeCorrection"/> 
+        <driver name="ReconClusterer" /> 
+        <driver name="CopyCluster" />
+
+        <!-- Hodoscope drivers -->
+        <driver name="HodoRunningPedestal"/>
+        <driver name="HodoRawConverter"/>
+
+        <!-- SVT reconstruction drivers -->
+        <driver name="RawTrackerHitSensorSetup"/>
+        <driver name="RawTrackerHitFitterDriver" />
+        <driver name="TrackerHitDriver"/>
+        <driver name="KalmanPatRecDriver"/>
+        <driver name="ReconParticleDriver_Kalman" /> 
+
+        <!--  DQM   -->
+        <!-- Following are optional Analysis Drivers  -->        
+        <driver name="EcalMonitoring"/>  
+        <driver name="SVTMonitoring"/>
+        <driver name="TrackingMonitoring"/>  
+        <driver name="FinalStateMonitoring"/>          
+        <driver name="V0Monitoring"/>          
+        <driver name="TridentMonitoring"/>
+	
+        <!-- Event filtering --> 
+        <!-- <driver name="StripEvent"/> --> 
+
+        <driver name="LCIOWriter"/>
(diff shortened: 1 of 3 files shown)


steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass2_recon.lcsim
```diff
@@ -0,0 +1,261 @@
...
+            <seedStrategy>0SBB000</seedStrategy>
+            <seedStrategy>000B0BA</seedStrategy>
+            <seedStrategy>ABSS000</seedStrategy>
+            <seedStrategy>SBB0000</seedStrategy>
+            <seedStrategy>SABS000</seedStrategy>
+	        <numPatRecIteration> 7 </numPatRecIteration>
+	        <numKalmanIteration> 0 </numKalmanIteration>
+	        <maxPtInverse> 8.881915381218574 </maxPtInverse>
+	        <maxD0> 39.249197341740356 </maxD0>
+	        <maxZ0> 8.37847994612359 </maxZ0>
```
mgignac, line 138: With the target at -7.5mm, where is the z0 distribution centered? Can we quickly check this isn't too tight? Plots should be produced as part of alignment runs.

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass2_recon.lcsim
```diff
@@ -0,0 +1,261 @@
...
+            <beamConMollerCandidatesColName>BeamspotConstrainedMollerCandidates_KF</beamConMollerCandidatesColName>
+            <beamConMollerVerticesColName>BeamspotConstrainedMollerVertices_KF</beamConMollerVerticesColName>
+            <targetConMollerCandidatesColName>TargetConstrainedMollerCandidates_KF</targetConMollerCandidatesColName>
+            <targetConMollerVerticesColName>TargetConstrainedMollerVertices_KF</targetConMollerVerticesColName>
+        </driver>  
+        <driver name="StripEvent" type="org.hps.recon.skims.MultiSkimDriver">
+	    <skimV0>true</skimV0>
+	    <v0ParamFile>v0skim_parameters_2019_ver0.txt</v0ParamFile>  
+	    <v0OutputFile>${outputFile}_v0skim.slcio</v0OutputFile>
+            <skimMoller>true</skimMoller>
```
mgignac, line 210: Lets remove the Mollers from output. Hopeless in 2019...

mgignac (changes requested):

steering-files/src/main/resources/org/hps/steering/recon/PhysicsRun2019_pass2_recon.lcsim
EBerzin: It's centered at ~0.2mm. I also did some test runs significantly loosening that cut, and it had no effect on the number of reconstructed tracks.

mgignac: Thanks. Track angles are quite small, even if the significant target offset... I figured it was close to zero, but just want to double check.

mgignac (approved):

mghrear (approved):

mghrear:
I was digging into this line-by-line to make the corresponding MC file, and I noticed something I wanted to run by both of you.

The beam variables under KalmanPatRecDriver are:
```
<beamPositionZ> 0.0 </beamPositionZ>
<beamSigmaZ> 0.02 </beamSigmaZ>
<beamPositionX> 0.0 </beamPositionX>
<beamSigmaX> 0.05 </beamSigmaX>
<beamPositionY> 0.0 </beamPositionY>
<beamSigmaY> 1.0 </beamSigmaY>
```
The beam variables under ReconParticleDriver_Kalman are:
```
<beamPositionX>0.200</beamPositionX>
<beamSigmaX>0.100</beamSigmaX>
<beamPositionY>0.02</beamPositionY>
<beamSigmaY>0.01</beamSigmaY>
<beamPositionZ>-7.5</beamPositionZ>
```
and the actual beam parameters are:
x: 0.11 +- 0.09
y: -0.01 +- 0.01
z: -7.54 +- 0.17

Are these differences meaningful / should we updated this?

mgignac:
Ah good catch! Yes, we should update this. For the 2021 dataset processing, are using a dynamic beam spot loaded through the conditions DB. In order for that to work, we will need this first pass so we can characterize the beam spot and then add it to the conditions DB. So for this first pass, we'll need to use a fixed one. 

@EBerzin , from your multi-track vertex fits, can you put in sensible (x,y) beam spot positions and uncertainties? The uncertainty on the 100 track fit will be quite a bit smaller than what we'd expect for 2-track vertex fits, so for the uncertainty I'd make it artificially large for the time being, OR we could use the values from 2021 which are probably pretty good (sigmaX 0.055 and sigmaY 0.045). It's not too important at this point. Forsure we need to make sure the vertex Z position is the same across both drivers.

@mghrear Note: for MC, we've typically generated beamspots at (x,y) = (0,0) until we have a better idea of how the data looks.

EBerzin:
I updated the beam positions and used the 2021 uncertainties.

mgignac:
Do we need a separate PR to get those changes in?

PR #1175, opened 2026-08-06
Kalman per-iteration hit timing cuts, plus SVT truth-relation diagnostics

mgignac:
Adds per-iteration control over the Kalman hit timing cuts, and the
diagnostics used to work out why tracks lose their truth relations under
pulser overlay.

### Kalman pattern recognition

`mxTdif` becomes a per-iteration array, joined by three new per-iteration
knobs:

- `seedTimeSpread` — gate on `tmax-tmin` across the 5 seed hits
- `hitTimeWindow` — pick up hits within a window of the running mean of
  accepted hit times, replacing the `tMin`/`tMax` envelope
- `maxTotalSpread` — backstop on total `tMax-tMin` over the track

All three default to 0, selecting existing behaviour, and
`setMaxTimeRange()` still sets both iterations, so existing steering files
are unaffected. `TrackCandidate.reFit()` gains a `trial` argument to
select the window; `removeHit()`/`addHits()` use the looser of the two
tiers since they have no iteration context.

Also exposes the iteration-1 seed cuts (`maxPtInverseIter1`, `maxD0Iter1`,
`maxZ0Iter1`, `maxTanLambdaIter1`) and `edgeTolerance` — the seed
detector-bounds check, previously hard-coded at 1 mm, which acts as an
acceptance cut rather than a quality cut at low tan(lambda). Seed-by-seed
tracing is armed per event from `debugEvent` instead of a compile-time
constant.

### SVT truth-relation diagnostics

A raw hit can lose its truth relation at either of two gates: G1
(`totalContrib > 4*meanNoise`) or G2 (`readoutCuts`, evaluated on the
merged waveform, whose baseline need not match the DB pedestal).
`SvtDigitizationWithPulserDataMergingReadoutDriver` now counts both,
indexed by whether the channel carried a pulser hit, and reports at
`endOfData()`. No behaviour change.

Opt-in `writeHitOriginCollections` (default false) writes
`SVTHitOriginPulser` and `SVTHitOriginMCContrib`, which together with
`SVTTrueHitRelations` resolve "no truth relation" into NOISE / MC_PURE /
MC_PURE_SUBTHRESH / PULSER_PURE / MERGED / MERGED_SUBTHRESH —
distinguishing a lost relation from an absent one. Enabling it widens the
output MCParticle collection, so output is bit-identical only with it off.

Two consumers: `SvtHitProvenanceDriver` (aggregate — are untruthed track
hits adjacent to MC-charged channels, against the event baseline?) and
`SvtEventForensicsDriver` (per-event hit dump separating "digitization
dropped the relation" from "patrec used different raw hits").

### Also

- Setter to control beamspot z sigma.
- `UnbiasedTriggerFilterDriver`: accept/reject counters and summary;
  pulser and FaradayCup split into separate branches for counting, same
  outcome.
- Removed a stray `</driver>` in the pass5 physics recon steering file.
- Regenerated `detector-model/dependency-reduced-pom.xml`.

### DIFF
--- a/analysis/src/main/java/org/hps/analysis/MC/SvtEventForensicsDriver.java
+++ b/analysis/src/main/java/org/hps/analysis/MC/SvtEventForensicsDriver.java
@@ -1,0 +1,397 @@
+package org.hps.analysis.MC;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import hep.physics.vec.Hep3Vector;
+
+import org.lcsim.event.EventHeader;
+import org.lcsim.event.LCRelation;
+import org.lcsim.event.MCParticle;
+import org.lcsim.event.RawTrackerHit;
+import org.lcsim.event.ReconstructedParticle;
+import org.lcsim.event.SimTrackerHit;
+import org.lcsim.event.Track;
+import org.lcsim.event.TrackState;
+import org.lcsim.event.TrackerHit;
+import org.lcsim.event.Vertex;
+import org.lcsim.util.Driver;
+
+/**
+ * Full per-hit dump of a handful of named events, for debugging a specific case where a
+ * real track lost its truth link under pulser overlay.
+ *
+ * SvtHitProvenanceDriver answers the aggregate question. This one answers the forensic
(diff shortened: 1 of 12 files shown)


bloodyyugo (commented):
This looks good to me.

EBerzin (approved):

bloodyyugo (approved):

PR #1179, opened 2026-08-14
Fix the KalTrack isolation shared hits sort (and downstream effects on MC hit matching etc)

bloodyyugo:
This PR fixes the issue of misidentifying reco tracks from made up of SimTrackerHit-matched hits as unmatched to an MCParticle.  

The issue in the code was from this snippet in KalTrack (hat-tip to Lauren for finding this):  

// KalTrack.java:533-546
List<Measurement> allHits = sensor.hits;              // NOT a copy
if (allHits.size() > 1 && ms.hitID > -1) {
    Measurement hitOnTrack = allHits.get(ms.hitID);   // read pre-sort: OK
    ...
    Collections.sort(allHits, Measurement.MeasurementComparatorUp);   // :546 mutates SiModule.hits
    int isoID = ms.hitID + awayFromBeam*nSteps;       // :550 stale index into the new order

... in words, the measurement list in the first line just points to the list on the SiModule and the sort then changes the order of these hits.  After this, the ms.hitID is basically meaningless.  

This bug effected a few collections after this snippet gets first run in KalmanPatRecDriver.java:  
KFTrackData isolations
KFGBLStripClusterData (potentially used for using KF in alignment)
KFUnbiasRes

### DIFF
--- a/recon/src/main/java/org/hps/recon/utils/Preselection2021.java
+++ b/recon/src/main/java/org/hps/recon/utils/Preselection2021.java
@@ -1,0 +1,893 @@
+package org.hps.recon.utils;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import hep.aida.IHistogram1D;
+import hep.aida.IHistogram2D;
+import hep.aida.IHistogramFactory;
+import hep.physics.vec.BasicHep3Vector;
+import hep.physics.vec.Hep3Vector;
+
+import org.hps.recon.ecal.cluster.ClusterUtilities;
+import org.hps.recon.tracking.CoordinateTransformations;
+import org.hps.recon.tracking.TrackData;
+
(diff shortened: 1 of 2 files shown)


bloodyyugo:
I also added Preselection2021 (a mockup of the hpstr preselection in hps-java) which I was using to debug the truth.

bloodyyugo:
Here are some plots proving this works...

The track purity for the electron in the V0 (can be from recoil): 
image

...and for positron: 
image

The V0 reco electron p vs. truth p from A' electron before bug-fix: 
Screenshot 2026-08-14 at 3 48 35 PM

...and after: 
Screenshot 2026-08-14 at 3 47 53 PM


Layer 0 isolation values (pre-bugfix peak at 0 is from code picking same hit as hit on track): 
image

mgignac (approved):

EBerzin (approved):

